PR #2, opened 2020-04-17
[WIP] First draft

jthorton:
## Description
This is the first draft of the qcsubmit rewrite based on the openforcefield toolkit.

The spec for the rewrite can be found here https://github.com/openforcefield/openforcefield/issues/516#issuecomment-588987644

## Todos
  - [ ] finish writing tests using snowflake qcarchive instance
 - [ ] functions to visualise the molecules in a file 
 - [ ] command line interface still needed 
- [x] more results containers 
 - [ ] have results auto collect when a calculation is submitted.
- [x] adding some find method to each workflow component that checks if the requirements are met 
- [ ] implement workflow modules which are predefined collections of components that allow for the quick creation of a workflow.
- [x] let results containers pull trajectories for optimisations and torsiondrives

## Questions
- [ ] Are there any other key filers/workflow components that should be implemented?
 - [ ] CLI packages that should be used?

## Status
- [ ] Ready to go


qcsubmit/results.py
```diff
@@ -0,0 +1,674 @@
...
+
+    @classmethod
+    def from_server(
+        cls,
+        client: ptl.FractalClient,
+        spec_name: str,
+        dataset_name: str,
+        include_trajectory: bool = False,
+        final_molecule_only: Optional[bool] = False,
+        subset: Optional[List[str]] = None,
```
mattwthompson, line 532: It would be nice to have a "only grab the first N records in this dataset" option to avoid `subset = list(opt_ds.data.records.keys())[:100]` that was needed in the example notebook. Not sure this is smart to have in the main class but maybe a util function somewhere, since it may only be useful for testing

jthorton: Good point am not sure how to introduce this though, maybe I could have a prepopulated list of 1000 keys stored in the package for testing?

mattwthompson: For now really all I need is the ability to grab N random records, which is something that we probably don't want accessible to the user in the main function. So I could just add that

For testing purposes, maybe that pre-populated list of keys that you know the behavior of? So you can know the expected behavior beforehand, usinng the same records each time, but not need to iterate through the entire dataset

Issue #3, opened 2020-05-14
Fields that should be required for OpenFF submissions


j-wags:
As discussed during the developers week, we would probably benefit by adding a requirement for metadata when submitting OpenFF jobs to QCArchive. The required metadata could include:
* `dataset.metadata[‘long_description’]` (basically the README from qca-dataset-submission at the time of submission)
* `dataset.metadata['long_description_url']` (point to qca-dataset-submission README for that submission)
* `dataset.metadata['submitter']`

Are there more required fields that we should add? We might start by thinking about what we _wish_ we had included in previous datasts @karmencj @hyejang @jthorton @ChayaSt @jchodera @mattwthompson @dotsdl @jmaat @trevorgokey

j-wags:
One idea that was floated in the QCFractal meeting today was to also have a 2D PDF created when submitting torsiondrives, highlighting the torsions that will be driven. This might be kinda hard to design, but could be good for human submission. For now I _wouldn't_ recommend that this become part of the default automation (since viewing an image would somehow add a GUI/image viewer to an "automated" workflow), but it might be a nice feature to think about i the long run.

jthorton:
This would be great to include as a receipt of the process of generating a dataset and is similar to what is currently made on submission using the scripts. Would the functions that depict the molecules in a pdf best live in the toolkit or here in some utility folder?

davidlmobley:
Ooh, yeah, I like the long description, long description URL, etc. Submitter, date, probably also a short description which is like a one-line description of what they are doing with it/the mnemonic they use to remember what dataset it is/why they ran it. Maybe called "dataset mnemonic".

As per call, metadata that can be filled automatically may be filled automatically, but not all data is in this category obviously.

trevorgokey:
For dataset submission, we would benefit from storing a set of symbols, and (charge, multiplicity) tuples in the dataset metadata. For starters, if we have the set of symbols, we can validate against the basis set we are using. If the basis set does not cover all of the symbols, we should reject submission. This is currently an issue for a recent set of molecules submitted containing iodine, but the "default" dzvp basis used does not cover them. All of these jobs failed, and they will never succeed since the qc_spec is incompatible with the molecule.

[...]

PR #7, opened 2020-05-19
Dataset metadata

jthorton:
## Description
Provide a brief description of the PR's purpose here.
addresses #3 
Metadata pydantic class is added to the datasets which holds important information about the submission.

## Todos
Autofill information
  - [X] Submitter
  - [X] Creation date
  - [X] Collection type
  - [X] Dataset name
  - [X] Short Description
  - [X] Long Description

User required fields
- [ ] URL

## Questions
- [ ] If all of the information which is normally included in the dataset-submission README (https://github.com/openforcefield/qca-dataset-submission/tree/master/2020-04-28-Fragmenter-test) is in the dataset would it make sense to have a function in the dataset which could automatically make the formatted README file?

- [X] Should any of those fields not be autogenerated? Note they are mutable so the user can easily swap them after generation.

- [X] The URL is currently an optional field which should be entered into the dataset, however, this could be added as an argument to the `create_dataset` function, for example

```python
from qcsubmit.factories import OptimizationDatasetFactory
opt_factory = OptimizationDatasetFactory()
# set settings here ...

opt_dataset = opt_factory.create_dataset(dataset_name="new dataset", molecules=molecules, description="testing qcsubmit", url="https://github.com/openforcefield/qcsubmit/")
```

edit
When creating a dataset from a factory a unique metadata field can be supplied straight to the dataset else it will be autogenerated as much as possible and users will have the ability to perform a final edit before submission.
```python
from qcsubmit.factories import OptimizationDatasetFactory
from qcsubmit.factories import Metadata
opt_factory = OptimizationDatasetFactory()
# set settings here ...
metadata = Metadata(submitter="jthorton", long_description_url="https://github.com/openforcefield/qcsubmit/" ...)
opt_dataset = opt_factory.create_dataset(dataset_name="new dataset", molecules=molecules, description="testing qcsubmit", metadata=metadata)
```

## Status
- [x] Ready to go


jthorton:
tagging @jaimergp for codecov help

PR #12, opened 2020-05-27
MM calculations

jthorton:
## Description
This PR allows new datasets made from the results of another to include cmiles in the molecule extras which is needed to accurately type the molecule when using MM.

## Todos
  - [x] Add tests to make sure the extras are added when requested.

## Questions
- [ ] For benchmarking we need to check what happens when a molecule is entered twice into the archive with only the extras being different.

## Status
- [ ] Ready to go

### DIFF
--- a/examples/Calculating collections with MM.ipynb
+++ b/examples/Calculating collections with MM.ipynb
@@ -1,0 +1,474 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This notebook will walk through the process of creating a new collection using QCSubmit and running MM clacultions such as optimizations and torsiondrives using QCEngine and QCFractal.\n",
+    "\n",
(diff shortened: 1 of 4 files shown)


trevorgokey:
I am not currently understanding why the cmiles needs to be stored in every molecule, since it is stored in the entry already. What is the reasoning here?

jthorton:
Yeah this is an awkward one that might have a better solution. It is to do with typing the molecule with the forcefield before running the MM calculation, using the cmiles we can accurately type the molecule, the trick is getting the data in the information that is sent to a worker as the entry metadata does not seem to pass along. I have found that sneaking it in the extras works see [here](https://github.com/MolSSI/QCEngine/pull/239) for the qcengine refactor which uses this.

jthorton:
This was fixed in the QCSpec update and all entries now have cmiles in the extras.

PR #25, opened 2020-06-25
Centralize validators for reuse

jthorton:
## Description
All validators are moved to one file so they can be shared between datasets and factories. All datasets now run validation when a molecule is entered into it or when loaded from a file instead of assuming the factory has handled validation.

This will form the basis of the automated workflow run on submissions to the qcarchive for validation.

Tests were also shortened by using fewer molecules.

##  Validators
- [X] dihedral connection for propers/impropers
- [X] linear torsion selection based on smarts
- [X] cmiles attributes 
- [X] scf properties
- [X] net charge warnings
- [X] detect molecular complexes

## Status
- [X] Ready to go


jthorton:
looks likes psi4 and the new openff package might not mix well.

jthorton:
Changing to the psi4 dev channel seems to of fixed this issue.

PR #34, opened 2020-08-03
Qcspec refactor

jthorton:
## Description
This allows factories and datasets to handle multiple QC specs which are heavily validated to avoid run time issues and can be submitted simultaneously. We also expand the qc spec to store wavefunction protocols allowing the saving of the wavefunction.

The second major change is that all datasets now have the cmiles stored at the molecule level, this is still hidden in the extras field but should allow running with MM methods in qcengine.

## Todos
  - [ ] It is not currently possible to supply the wavefunction protocol to standard datasets we can use a workaround or wait till this [PR](https://github.com/MolSSI/QCFractal/pull/608) is merged. Probably use the workaround and note to update QCSubmit when this is fixed in fractal.

## Status
- [ ] Ready to go

### DIFF
--- a/qcsubmit/common_structures.py
+++ b/qcsubmit/common_structures.py
@@ -8,6 +8,7 @@ from typing import Any, Dict, List, Optional, Set, Tuple, Union
 
 import numpy as np
 import qcportal as ptl
+from qcelemental.models.results import WavefunctionProtocolEnum
 from pydantic import BaseModel, HttpUrl, constr, validator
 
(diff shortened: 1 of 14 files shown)


dotsdl:
You're good to go to clear out your workaround; MolSSI/QCFractal#608 is now merged! Thank you @jthorton!

jthorton:
Thanks for the heads up Ill strip this once we have a new release

PR #37, opened 2020-09-22
RMSDCutoffConformerFilter custom workflow component

trevorgokey:
## Description
Adds a filter that prunes the number of conformers such that all pairwise RMSD are greater than some cutoff.

## Todos
  - [x] Adds the `RMSDCutoffConformerFilter` custom workflow component

## Status
- [x] Ready to go

### DIFF
--- a/qcsubmit/workflow_components/__init__.py
+++ b/qcsubmit/workflow_components/__init__.py
@@ -4,6 +4,7 @@ from .filters import (
     CoverageFilter,
     ElementFilter,
     MolecularWeightFilter,
+    RMSDCutoffConformerFilter,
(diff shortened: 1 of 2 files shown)


jthorton:
@trevorgokey Thanks for the PR I think this would be a useful for conformer generation but maybe it would be better to use the toolkit for this if possible. The toolkit conformer generation API was expanded to accept a `rms_cutoff` setting for both RDKit and Openeye and is currently used as the default is 1 angstrom.

trevorgokey:
That might be something to look at then.

Because another PR encloses these commits, I am closing and deleting this one. Thanks for the input!

PR #39, opened 2020-09-22
Prevent double-adding a stereoisomer in EnumerateStereoisomers

### LINKED ISSUE #38: EnumerateStereoisomers adds same molecule twice if isomers are found and include_input=True
Using the EnumerateStereoisomers functionally allows the original molecule to be included:
```
        for molecule in molecules:
            try:
                isomers = molecule.enumerate_stereoisomers(
                    undefined_only=self.undefined_only,
                    max_isomers=self.max_isomers,
                    rationalise=self.rationalise,
                    toolkit_registry=toolkit,
                )
                if self.include_input:
                    result.add_molecule(molecule)

                for isomer in isomers:
                    result.add_molecule(isomer)
```
It appears that:
1. If no isomers are found and include_input=False, then no molecules are produced from this component. Does this mean there is no output from this component for downstream components? 
2. If a chiral center is found and isomers=2 and include_input=True, then 3 molecules are added. I am assuming the input here is identical to one of the stereoisomers.

trevorgokey:
## Description
Fixes #38. It appears that we should only add the original molecule if either 1. no stereoisomers are found or 2. If the setting is explicitly set to True. This PR changes the default setting of `include_input` to False.

## Todos
- [x] Fixes #38

## Status
- [x] Ready to go

### DIFF
--- a/qcsubmit/workflow_components/__init__.py
+++ b/qcsubmit/workflow_components/__init__.py
@@ -4,6 +4,7 @@ from .filters import (
     CoverageFilter,
     ElementFilter,
     MolecularWeightFilter,
+    RMSDCutoffConformerFilter,
     RotorFilter,
     SmartsFilter,
 )
--- a/qcsubmit/workflow_components/filters.py
+++ b/qcsubmit/workflow_components/filters.py
@@ -2,6 +2,7 @@
 File containing the filters workflow components.
 """
 import re
+import numpy as np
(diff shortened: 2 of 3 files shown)


trevorgokey:
I realize there is another layer here that was not obvious, which was the deduplication scheme that occurred in the background.

trevorgokey:
This is a subset of #40, so closing

PR #40, opened 2020-09-23
Components with multiprocessing

trevorgokey:
## Description
Implements multiprocessing in the component workflow and a few other items to improve performance. This also takes a first step in defining ways to add configuration of the workflow components, but should not be part of the model.

## Todos
  - [x] Computationally expensive tasks such as conformer generation are much faster
  - [x] `RMSDCutoffConformerFilter`
  - [x] Compression and decompression support
  - [x] Implements architecture to handle private members in the `CustomWorkFlowComponent` Pydantic models
  - [x] The `_apply_init`, `_apply`, and `_apply_finalize` methods in `CustomWorkFlowComponent`  for multiprocessing support
 - [x]  Introduces `FilterComponent` and `StateEnumeratorComponent` base classes to help manage configuration
  - [x] Exposes the `runtime_set_processes` and `runtime_set_skip_unique_check` methods in `CustomWorkFlowComponent`
  - [x] Adds an option to create searchable text inside the PDFs created by `visualize` (RD only for now)

## Questions
- [x] Deduplication occurs often and I provided a `skip_unique_check` in `ComponentResult`. What are the drawbacks of this?
    Deduplication speed has been dramatically improved, but `skip_unique_check` has been kept

## Status
- [ ] Ready to go


jthorton:
@trevorgokey Thanks for the PR this will very useful for QCSubmit as the performance improvement is fantastic! 
Just going to update here with our slack conversion points.
- As you pointed out deduplication is very slow and is done a lot and does not make sense after every component as some will never produce new molecules. 

I have now added a [new deduplication method](https://github.com/openforcefield/qcsubmit/pull/41) which should make the step negligible compared to workflow component, please can you test this out to make sure it improves the timings you were seeing and make sure you still see the same deduplication of molecules? If we still need to remove deduplication from every step then I think we could add a private variable which indicates if the component produces duplicates and this will stop the deduplication process like you have but this should not be changeable by the user.   

- > fancier would be having the base class ask for the number of processors at runtime, preventing the ability to change it prior to calling apply 

I agree I think so that we can set the number of available processes at the factory level once when creating the dataset the `apply` method in the component base class should be changed to take processors variable rather than setting it for each component. 

In the case, we find a component that should not be run in parallel I would suggest we have other private settings variable like `process_parallel` which then determines if the component can be run in parallel regardless of the number of processors passed the apply function. 

Maybe to easily allow the extension to future private settings variables we could add a `_settings` (This is then hidden and is not stored in the json output) attribute in the base class which accepts a `Settings` immutable pydantic class which is set in the constructor something like this 
```python
class Settings(BaseModel):
    process_parallel: bool = True
    produces_duplicates: bool = True

    class Config:
        allow_mutation: bool = False
         extra: "forbid"

```

trevorgokey:
I think it might be useful to open a new feature issue to discuss how mutable/immutable settings can be managed. This PR is ready to go, but if it depends on some general architecture design changes we should try to split that off from here and chew on it separately.

jthorton:
Good point Ill open a new one for the settings, for now then can we move the processors to be an argument in the `apply` method rather than being set per component and then have the factories pass there new `processors` parameter in the apply call when creating datasets?

qcsubmit/datasets.py
```diff
@@ -227,10 +233,11 @@ def add_molecule(self, molecule: off.Molecule):
                         )
             else:
                 # molecule already in list and coords not present so just return
-                return
+                return True
```
jthorton, line 236: Do you still need this or was this just for testing? If we need it please change the type hint to reflect this.

qcsubmit/workflow_components/base_component.py
```diff
@@ -18,11 +19,17 @@ class should allow users to easily create new work flow components with out need
     component_name: str
     component_description: str
     component_fail_message: str
+    _processes: Union[None, int] = 1  # By default everything is done in serial
```
jthorton, line 22: we can remove this if we are passing the number via the run function.

qcsubmit/workflow_components/base_component.py
```diff
@@ -61,7 +139,7 @@ def provenance(self) -> Dict:
         """
         ...
 
-    def _create_result(self) -> ComponentResult:
+    def _create_result(self, **kwargs) -> ComponentResult:
```
jthorton, line 142: Good idea this will be helpful.

qcsubmit/workflow_components/conformer_generation.py
```diff
@@ -22,42 +22,48 @@ class StandardConformerGenerator(ToolkitValidator, CustomWorkflowComponent):
     component_fail_message = "Conformers could not be generated"
 
     # custom components for this class
+    rms_cutoff: Union[None, float] = None
```
jthorton, line 25: Thanks for adding this in as well.

qcsubmit/workflow_components/filters.py
```diff
@@ -56,15 +60,15 @@ def apply(self, molecules: List[Molecule]) -> ComponentResult:
 
         from rdkit.Chem import Descriptors
 
-        result = self._create_result()
-        for molecule in molecules:
+        result = self._create_result
```
jthorton, line 63: I think we are just missing some brackets here.

qcsubmit/workflow_components/filters.py
```diff
@@ -134,6 +138,10 @@ class ElementFilter(BasicSettings, CustomWorkflowComponent):
         "I",
     ]
 
+    cache: Union[List[int], None] = None
```
jthorton, line 141: Lets make this a private parameter in the base class so all components can use it and it won't make its way into the json that way. We can also make the base class `apply_finalize` function clear the cache.

qcsubmit/workflow_components/state_enumeration.py
```diff
@@ -169,30 +204,27 @@ def apply(self, molecules: List[Molecule]) -> ComponentResult:
...
                 try:
                     protomers = molecule.enumerate_protomers(max_states=self.max_states)
 
-                    for protomer in protomers:
-                        result.add_molecule(protomer)
-                    result.add_molecule(molecule)
-
-                except Exception:
-                    self.fail_molecule(molecule=molecule, component_result=result)
+                    if len(protomers) == 0:
```
jthorton, line 218: we need to be careful here as we may lose the input protomer as the toolkit removes them from the return I think.

qcsubmit/workflow_components/state_enumeration.py
```diff
@@ -169,30 +204,27 @@ def apply(self, molecules: List[Molecule]) -> ComponentResult:
...
+                        result.add_molecule(molecule)
 
-            return result
+                    else:
+                        for protomer in protomers:
```
jthorton, line 222: we need to be careful here as we may lose the input protomer as the toolkit removes them from the return.

qcsubmit/workflow_components/state_enumeration.py
```diff
@@ -169,30 +204,27 @@ def apply(self, molecules: List[Molecule]) -> ComponentResult:
...
+                            result.add_molecule(protomer)
 
+                except Exception:
+                    result.filter_molecule(molecule)
         else:
+            result.filtered = molecules
```
jthorton, line 228: This needs to be changed to reflect the dict used in the new deduplication method.

qcsubmit/workflow_components/state_enumeration.py
```diff
@@ -169,30 +204,27 @@ def apply(self, molecules: List[Molecule]) -> ComponentResult:
             This is only possible using Openeye so far, if openeye is not available this step will fail.
         """
 
-        from openforcefield.utils.toolkits import OpenEyeToolkitWrapper
+        result = self._create_result(skip_unique_check=self.skip_unique_check)
 
-        result = self._create_result()
+        has_oe = self.cache
```
jthorton, line 209: Not sure if this is needed as the factory should raise an error if you add this component to a workflow but do not have openeye installed so we will never hit the apply method. The only time this would be needed is if the component is used in isolation but I am not sure how common that use case is.

trevorgokey: I am actually thinking that we should not actually make a hard requirement on OE, as that type of detail is an OpenFF TK implementation detail that we should not need to consider.

qcsubmit/workflow_components/state_enumeration.py
trevorgokey: Submitted https://github.com/openforcefield/openforcefield/issues/738 to see if we can get some blanket clarification in the TK

jthorton: as long as it hasn't been changed since I put the state enumeration methods in the toolkit they all should remove the input molecule regardless of the backend toolkit used see here for the molecule class docstrings https://github.com/openforcefield/openforcefield/blob/47d98f49109338ec1c29b907f98ace4d1fdbbb53/openforcefield/topology/molecule.py#L3947.

qcsubmit/workflow_components/state_enumeration.py
jthorton: I agree to some extent, the current method will raise an error when you create the workflow about openeye being missing before creating the dataset rather than running the whole workflow and finding all molecules have failed a component that has the missing dependency which might cause users to think there is a problem with the workflow rather than the environment.

qcsubmit/workflow_components/state_enumeration.py
trevorgokey: Ah! I was careless and only read the main docstring. Didn't think I had to read the return data description. 
That said, though, I do think we are getting the "same" input back. Going to carry the conversation on to #738

qcsubmit/workflow_components/state_enumeration.py
trevorgokey: So in this case, we should be catching these types of environment things in either the constructors or the `add_workflow_component`. I agree that these errors would be confusing once the molecules are being processed.

qcsubmit/datasets.py
trevorgokey: Was for testing, but I don't see the harm in leaving since it lets me know if the molecule was deduplicated

qcsubmit/workflow_components/filters.py
trevorgokey: Done!

qcsubmit/workflow_components/state_enumeration.py
trevorgokey: Done!

jthorton:
This will also fix #42

trevorgokey:
Alright I think I've added my last little bit before I set this PR down. The searchable PDF has some very nice functionality, but is hard to get right if one wants to make it look "pretty" on a per-page basis. The point now is that there is some code that 1. embeds actual text in the PDF that can be searched, and 2. does the entire thing in memory without touching any files or slowdown from disk access. Comes at a cost though, as it needs `wkhtmltopdf` from conda, and the light wrapper `pdfkit` from `pip`. The functionality is exposed via `visualize(..., annotate=True)` and uses the old method if it is `False`. Does nothing in OE currently.

Also, I found that my initial issues with the RD version was that the image was just too large and would crash (it wanted to put 10000 molecules on a single pdf page). In the future this can be broken up into smaller pages using the PIL saver; I used it with success before switching to searchable text with html->pdf.

trevorgokey:
Also, let me know if there is anything I should be doing regarding the tests

jthorton:
Thanks for changing the RDKit bit to another branch the rest looks great thanks for all the hard work! Ill fix the tests it looks like the master merge might have messed things up.

jthorton (approved):

PR #44, opened 2020-10-01
Added install instructions for operating without OpenEye dependency

dotsdl:
## Description
Adds install instructions for operating without the OpenEye Toolkit.

## Status
- [x] Ready to go


jthorton:
@dotsdl Thanks for updating this but unfortunately I could not get that environment to resolve without adding the openeye channel as fragmenter has openeye in its recipe as well. Ill work on making fragmenter optional and update the recipe.

dotsdl:
Sounds good @jthorton, thanks! Meanwhile @j-wags and @mattwthompson are planning to put de-openeyeing fragmenter on the docket for next week.

PR #51, opened 2020-10-13
Split RDKit pdfs

### LINKED ISSUE #50: RDKit crashes when visualizing large datasets
Using the RDKit backend to visualize large datasets via PDF crashes as a single page PDF is attempted to be made. We need to find a way to split the PDF into multiple pages. Maybe rendering a new pdf every so many molecules and then combining them together using some package. The [pypdf2](https://github.com/mstamy2/PyPDF2) pure python package could be a good option and it is on conda-forge.

jthorton:
## Description
This PR should fix #50 by splitting the dataset into 24 molecule chunks these are then visualized into separate PDFs which are merged back together using pypdf2. 

## Status
- [ ] Ready to go

### DIFF
--- a/devtools/conda-envs/psi4_env.yaml
+++ b/devtools/conda-envs/psi4_env.yaml
@@ -27,6 +27,7 @@ dependencies:
   - qcfractal >=0.13.0
   - torsiondrive
   - fragmenter
+  - pypdf2
 
     # Pip-only installs
   - pip:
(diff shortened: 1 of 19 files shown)


jthorton:
PDFs made using RDKit however can not be searched like those from Openeye.

trevorgokey:
The implementation in this PR can be accomplished via standard PIL options [1]. If we have a list of PDF outputs from RDkit, e.g. `images = [grid1, grid2, grid3]`, we can use
```
images[0].save("out.pdf", append_images=images[1:], save_all=True)
```

[1] https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#pdf

jthorton:
@trevorgokey thanks for the spotting this it works great!

PR #74, opened 2020-12-15
[WIP] Update to rotor filter component

pavankum:
Added functionality to filter rotors by minimum no. of rotors to be present, additional `filter_by` arg in the function which defaults to 'max' so that previous tests and functionality are unaffected, and added new tests

## Description
If we decide to fragment molecules during dataset creation that have rotors above a threshold value this filter helps in keeping only molecules that meet this threshold. 

## Todos
  - [x] Filters rotors by minimum number of rotors to be present

## Status
- [ ] Ready to go

### DIFF
--- a/qcsubmit/workflow_components/filters.py
+++ b/qcsubmit/workflow_components/filters.py
@@ -357,14 +357,18 @@ class RotorFilter(BasicSettings, CustomWorkflowComponent):
     """
 
     component_name = "RotorFilter"
-    component_description = "Filter the molecules based on the minimum/maximum number of allowed rotatable bonds."
+    component_description = "Filter the molecules based on the minimum or/and maximum number of allowed " \
+                            "rotatable bonds. When the filter is applied without supplying any values for " \
+                            "maximum_rotors or minimum_rotors then it defaults to at least 1 rotor and upto " \
+                            "a maximum of 4 rotors. None can also be supplied to apply only an upper or" \
+                            "a lower bound on the number of allowed rotors."
     component_fail_message = "The molecule has too many(low) rotatable bonds."
-    minimum_rotors: int = Field(
+    minimum_rotors: Optional[int] = Field(
         1,
         description="The minimum number of rotatable bonds to be present in the molecule.",
     )
 
-    maximum_rotors: int = Field(
+    maximum_rotors: Optional[int] = Field(
         4, description="The maximum number of rotatable bonds allowed in the molecule."
     )
     _properties = ComponentProperties(process_parallel=True, produces_duplicates=False)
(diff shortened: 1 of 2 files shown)


jthorton:
@pavankum Thanks for putting this together this looks great. One thing I think could be changed is the `filter_by` argument this would then let users filter by both max and min at the same time in one go like the element/weight filters. Users could still filter by max/min if we allow the other to take `None` which would skip the check so we would have something like

```python
# filter molecules with between 2 and 6 rotatable bonds
rotor_filter.maximum_rotors = 6 
rotor_filter.minimum_rotors = 2

# filter molecules with at least 2 rotatable bonds
rotor_filter.maximum_rotors = None
rotor_filter.minimum_rotors = 2

# filter molecules with no more than 5 rotatable bonds
rotor_filter.maximum_rotors = 5
rotor_filter.minimum_rotors = None
```
otherwise users wishing to filter by max and min in the workflow would have to add the component twice which is possible but more work.

pavankum:
@jthorton Thanks Josh for the review. Yeah, it makes sense to have a general filter, I made the changes.

qcsubmit/workflow_components/filters.py
```diff
@@ -349,18 +349,20 @@ def provenance(self) -> Dict:
...
         Rotatable bonds are non terminal torsions found using the `find_rotatable_bonds` method of the
         openforcefield.topology.Molecule class.
     """
 
     component_name = "RotorFilter"
-    component_description = (
-        "Filter the molecules based on the maximum number of allowed rotatable bonds."
+    component_description = "Filter the molecules based on the minimum/maximum number of allowed rotatable bonds."
+    component_fail_message = "The molecule has too many(low) rotatable bonds."
+    minimum_rotors: int = Field(
```
jthorton, line 362: This should now have the type `Optional[int]` to show that `None` can be supplied. I think we should also explain what happens if `None` is supplied in the field description.

qcsubmit/workflow_components/filters.py
```diff
@@ -385,9 +387,11 @@ def _apply(self, molecules: List[Molecule]) -> ComponentResult:
 
         # run the the molecules and calculate the number of rotatable bonds
         for molecule in molecules:
-            if len(molecule.find_rotatable_bonds()) > self.maximum_rotors:
+            if (
```
jthorton, line 390: We need some logic to handle the case where `maximum_rotors` or `minimum_rotors` is `None`. Lets also only call `find_rotatable_bonds` once and save the result for the comparison.

jthorton:
@pavankum thanks for making the changes only the last two things to change and this should be good to go.

qcsubmit/workflow_components/filters.py
pavankum: > @pavankum thanks for making the changes only the last two things to change and this should be good to go.

@JoshHorton My apologies for the delay in fixing these issues. Please review the changes when you have time. Thank you.

jthorton:
@pavankum Thanks for starting this I have finished this in #132 as this has been parked for a while let me know if you find any issues.

PR #82, opened 2021-01-12
Smirnofffrosst fix

### LINKED ISSUE #79: SmirnoffFrosst validation error
Currently, smirnoffFrosst forcefields can not be used in production as they fail validation due to the capitalization in their name.

jthorton:
## Description
This will fix #79 allowing the smirnoff99Frosst forcefields to be correctly validated. I have also added one of these forcefields to the test submissions section to make sure this does not break in future.

## Status
- [ ] Ready to go


jthorton:
For some reason, the tests seem to behave differently between mac and Linux, supplying the forcefield method as `smirnoff99frosst-1.0.0` works on mac but raises an error on Linux due to the name being lowered not quite sure if I can fix this here or if this should be fixed upstream in the maybe.

PR #103, opened 2021-03-20
Fix enumeration

jthorton:
## Description
This PR fixes tautomer enumeration which was dropping the input molecules meaning only the enumerated forms were kept.

## Todos
  - [X] Add tests to make sure input molecules are kept when enumerating states.

## Questions
- [ ] Question1

## Status
- [x] Ready to go


SimonBoothroyd:
Thanks for this @jthorton - great catch!

PR #105, opened 2021-04-05
Create CMILES object directly from a molecule

SimonBoothroyd:
## Description

This PR adds a convenience method to create a CMILES containing `MoleculeAttributes` object from an OpenFF molecule. This allows the object to be created directly without first needing to create a dataset factory first.

## Questions
- [x] I named this method `from_openff_molecule` to make it explicit that this method creates from an OpenFF molecule rather than say a QCElemental molecule, but it may be better to just have a `from_molecule` method and handle the molecule type in the method itself, maybe making private `_from_openff_molecule`, `_from_qcel_molecule`, `_from_XXX_molecule` methods to handle the different molecule cases. 

## Status
- [ ] Ready to go

### DIFF
--- a/openff/qcsubmit/common_structures.py
+++ b/openff/qcsubmit/common_structures.py
@@ -1038,6 +1038,52 @@ class MoleculeAttributes(DatasetConfig):
         ..., description="The standard inchi key given by the inchi program."
     )
 
+    @classmethod
+    def from_openff_molecule(cls, molecule) -> "MoleculeAttributes":
+        """Create the Cmiles metadata for an OpenFF molecule object.
+
+        Parameters:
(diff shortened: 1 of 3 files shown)


jthorton:
@SimonBoothroyd thanks for putting this together this is great and helps with the awkwardness of using a factory to make the CMILS objects. WRT to the naming let's stick with `from_openff_molecule` for now I think only an openff molecule is safe to use as input due to missing information in the likes of qcelemental molecules and I like the idea of making it very clear that we are using a openff molecule as our goto structure which users might not expect when working with qcfractal. 

Please feel free to merge tests are broken due to fragmenter and an rdkit update which I will work on separately.

jthorton (approved):

PR #113, opened 2021-04-23
Add cached versions of QC record and molecule queries

SimonBoothroyd:
## Description

This PR adds a number of helper methods which enable caching queries to a `QCFractal` instance, as well as retrieving the records associated with a result collection.

These changes will likely be superseded by https://github.com/MolSSI/QCFractal/pull/671 (or its successors) when ready.

## Status
- [X] Ready to go

### DIFF
--- a/openff/qcsubmit/results/filters.py
+++ b/openff/qcsubmit/results/filters.py
@@ -2,8 +2,13 @@ import abc
 import logging
 from typing import TYPE_CHECKING, List, Optional, TypeVar
 
+import numpy as np
 from openff.toolkit.topology import Molecule
 from pydantic import BaseModel, Field, root_validator
+from qcportal.models import Molecule as QCMolecule
+from qcportal.models import OptimizationRecord, ResultRecord, TorsionDriveRecord
+from qcportal.models.records import RecordBase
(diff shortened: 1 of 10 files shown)


openff/qcsubmit/results/filters.py
```diff
@@ -19,6 +24,112 @@ class ResultFilter(BaseModel, abc.ABC):
...
+        Returns:
+            The record mapped to an OpenFF molecule.
+        """
+
+        molecule: Molecule = Molecule.from_mapped_smiles(
+            cmiles, allow_undefined_stereo=True
+        )
+        qc_molecule: QCMolecule
+
+        for qc_molecule in record.get_final_molecules().values():
```
jthorton, line 99: For the fitting, we need the molecules and the results and calling these functions on every record can be quite slow, maybe we should batch the calls, so collect together the ids of the molecules and results we want and query them in chunks?

SimonBoothroyd: Yeah that's a really good point! I've reconsidered this PR and now gone with a fully cached solution. I think really this is the only clean way to ensure we don't hit these expensive query calls.

PR #115, opened 2021-04-26
Make `openmmforcefields` an optional dependency

SimonBoothroyd:
## Description
This PR make `openmmforcefields` an optional dependency. This removes the second-hand `ambertools` dependency of this package and makes it easier to install alongside `psi4` when GAFF isn't needed.

## Status
- [X] Ready to go


jthorton (approved):
Great idea I love making the package smaller when we can!

PR #120, opened 2021-04-27
Add record status result filter

SimonBoothroyd:
## Description

This PR adds a `RecordStatusFilter` filter which will only retain records whose status matches a specified value.

## Status
- [X] Ready to go


jthorton (approved):
LGTM!

PR #121, opened 2021-04-27
Implement optimization collection to basic dataset

SimonBoothroyd:
## Description

This PR implements the optimization collection `create_basic_dataset` function which will create a `BasicDataset` from an existing `OptimizationResultCollection`. This should allow hessian datasets to easily be created from existing optimization results.

## Status
- [X] Ready to go


openff/qcsubmit/results/results.py
```diff
@@ -486,7 +486,48 @@ def create_basic_dataset(
...
-        raise NotImplementedError()
+
+        records = self.to_records()
+
+        dataset = BasicDataset(
+            dataset_name=dataset_name,
+            description=description,
+            dataset_tagline=tagline,
+            driver=driver,
+            metadata={} if metadata is None else metadata,
```
jthorton, line 497: Should we be using the same metadata as the original input dataset or be making a new one? I guess this is further complicated if we are combining multiple datasets together.

openff/qcsubmit/results/results.py
```diff
@@ -486,7 +486,48 @@ def create_basic_dataset(
...
+                ),
+                molecule=molecule,
+                attributes=MoleculeAttributes.from_openff_molecule(molecule),
+                extras=record.extras,
+                keywords=record.keywords,
+            )
+
+        dataset.clear_qcspecs()
+
+        for i, (method, basis, program) in enumerate(qc_specs):
```
jthorton, line 520: I wonder if it would be better to make a dataset with no `qc_specs` this way the user has to decide or they will get an error at submission, this helps with the awkward case where a user has combined two optimisation datasets with different specifications which will result in the full hessian dataset being run with both which might not be expected?

openff/qcsubmit/results/results.py
SimonBoothroyd: Yeah I wasn't sure about this one either because of the fact that the records may have come from different datasets to begin with. 

I think making the user re-populate this is probably the safest way to go as then they'll need to explicitly think through what they are about to do, rather than something surprising happening by default!

openff/qcsubmit/results/results.py
SimonBoothroyd: Yeah that's a really good point and I definitely agree - I think in general it probably makes sense for these methods to only populate the entries of the dataset and nothing else (including the metadata) for the reasons you state.

If they do want to just use the same specs from the 'original' dataset(s) then that should only be 1-2 lines extra that they'd need to add to their scripts I think!

SimonBoothroyd: I partially made this change in `1fe1b32` - I wasn't sure whether we should add no computes specs to the dataset or the default if the user hasn't specified?

jthorton: Ah good point, I think for our use case the default spec should be fine.

PR #123, opened 2021-04-27
Add fixed hydrogen inchi/inchikey to attributes

jthorton:
## Description
We often use fixed hydrogen inchi key for fast deduplication, this PR adds the fixed hydrogen inchi and inchikey to the `MoleculeAttributes` object, these fields are automatically generated still via the `from_openff_molecule` classmethod. A convenient method to generate an openff molecule from the MoleculeAttributes is also added called `to_openff_molecule`.


SimonBoothroyd (approved):
LGTM - the addition of `to_openff_molecule` seems like it will come in handy a lot!

PR #125, opened 2021-04-28
Charge result filter

jthorton:
## Description
This PR adds a charge result filter.

## Status
- [X] Ready to go

### DIFF
--- a/openff/qcsubmit/results/filters.py
+++ b/openff/qcsubmit/results/filters.py
@@ -384,18 +384,18 @@ class RecordStatusFilter(ResultRecordFilter):
 
 
 class ChargeFilter(CMILESResultFilter):
-    """A filter which will only retain records if their charge matches allowd values or is not in the
+    """A filter which will only retain records if their formal charge matches allowed values or is not in the
     exclude list."""
 
     charges_to_include: Optional[List[int]] = Field(
         None,
-        description="Only molecules with a net charge in this list will be kept. "
+        description="Only molecules with a net formal charge in this list will be kept. "
         "This option is mutually exclusive with ``charges_to_exclude``.",
     )
 
     charges_to_exclude: Optional[List[int]] = Field(
         None,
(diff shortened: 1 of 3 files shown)


openff/qcsubmit/results/filters.py
```diff
@@ -381,3 +381,48 @@ def _filter_function(
     ) -> bool:
 
         return record.status.value.upper() == self.status.value.upper()
+
+
+class ChargeFilter(CMILESResultFilter):
+    """A filter which will only retain records if their charge matches allowd values or is not in the
```
SimonBoothroyd, line 387:
```suggestion
    """A filter which will only retain records if their formal charge matches allowed values or is not in the
```

openff/qcsubmit/results/filters.py
```diff
@@ -381,3 +381,48 @@ def _filter_function(
...
+    exclude list."""
+
+    charges_to_include: Optional[List[int]] = Field(
+        None,
+        description="Only molecules with a net charge in this list will be kept. "
```
SimonBoothroyd, line 392:
```suggestion
        description="Only molecules with a net formal charge in this list will be kept. "
```

openff/qcsubmit/results/filters.py
```diff
@@ -381,3 +381,48 @@ def _filter_function(
...
+        "This option is mutually exclusive with ``charges_to_exclude``.",
+    )
+
+    charges_to_exclude: Optional[List[int]] = Field(
+        None,
+        description="Any molecules with a net charge which matches any of these values will be removed. "
```
SimonBoothroyd, line 398:
```suggestion
        description="Any molecules with a net formal charge which matches any of these values will be removed. "
```

SimonBoothroyd (approved):
This looks great! There were just a few minor typos and maybe some clarifications that could be added but otherwise LGTM!

PR #126, opened 2021-04-29
Add debug logging to result querying

SimonBoothroyd:
## Description

This PR adds debug logging calls to the cached query methods. This will allow users to see which step of the query pipeline is currently executing should they wish.

## Status
- [X] Ready to go

### DIFF
--- a/openff/qcsubmit/results/caching.py
+++ b/openff/qcsubmit/results/caching.py
@@ -1,5 +1,6 @@
 """A module which contains helpers for retrieving data from QCA and caching the
 results in memory for future requests."""
+import logging
 from functools import lru_cache
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, TypeVar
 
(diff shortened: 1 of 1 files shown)


openff/qcsubmit/results/caching.py
```diff
@@ -101,7 +104,15 @@ def _cached_client_query(
 
     client = _cached_fractal_client(client_address)
 
-    for batch_ids in _batched_indices(missing_query_ids, client.query_limit):
+    logger.debug(f"starting {query_name} to {client_address}")
+
+    batch_query_ids = _batched_indices(missing_query_ids, client.query_limit)
+    logger.debug(f"query split into {len(batch_query_ids)} batches")
+
+    for i, batch_ids in enumerate(batch_query_ids):
```
jthorton, line 112: A good place to use tqdm to show the progress of the queries.

jthorton (approved):
Hey this looks good as a first pass at logging, happy to have this in and we can come back to adding tqdm progress bars to the batch calls later if you like.

openff/qcsubmit/results/caching.py
SimonBoothroyd: Definitely agree!

SimonBoothroyd:
> happy to have this in and we can come back to adding tqdm progress bars to the batch calls later if you like.

Thanks - I think I'll leave the progress bars for a separate PR to this one nice and light!

PR #127, opened 2021-04-29
Results Element filter

jthorton:
## Description
This PR adds a results element filter.

## Todos
  - [X] Make the element validation function a general validator that can be reused.

## Status
- [X] Ready to go

### DIFF
--- a/openff/qcsubmit/validators.py
+++ b/openff/qcsubmit/validators.py
@@ -275,11 +275,10 @@ def check_allowed_elements(element: Union[str, int]) -> Union[str, int]:
 
     if isinstance(element, int):
         return element
-    else:
-        try:
-            _ = Element.getBySymbol(element)
-            return element
-        except KeyError:
-            raise ValueError(
-                f"An element could not be determined from symbol {element}, please enter symbols only."
-            )
+    try:
+        _ = Element.getBySymbol(element)
+        return element
+    except KeyError:
+        raise ValueError(
+            f"An element could not be determined from symbol {element}, please enter symbols only."
+        )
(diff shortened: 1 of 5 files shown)


openff/qcsubmit/validators.py
```diff
@@ -259,3 +259,27 @@ def check_valence_connectivity(molecule: qcel.models.Molecule) -> qcel.models.Mo
         )
 
     return molecule
+
+
+def check_allowed_elements(element: Union[str, int]) -> Union[str, int]:
```
SimonBoothroyd, line 264: This is great - good catch that this would be duplicate code!

openff/qcsubmit/validators.py
```diff
@@ -259,3 +259,27 @@ def check_valence_connectivity(molecule: qcel.models.Molecule) -> qcel.models.Mo
...
+    if isinstance(element, int):
+        return element
+    else:
+        try:
+            _ = Element.getBySymbol(element)
+            return element
+        except KeyError:
+            raise ValueError(
+                f"An element could not be determined from symbol {element}, please enter symbols only."
+            )
```
SimonBoothroyd, line 285:
```suggestion
    try:
        _ = Element.getBySymbol(element)
        return element
    except KeyError:
        raise ValueError(
            f"An element could not be determined from symbol {element}, please enter symbols only."
        )
```

We can get away without the `else` statement here as the `if` statement returns when it is triggered. This lets us un-nest the larger code block which usually helps with readability 🙂

SimonBoothroyd (approved):
This is a great addition! I noticed just a few little nit-picky lint changes but otherwise this LGTM!

openff/qcsubmit/validators.py
jthorton: Ah, good spot thanks!

PR #129, opened 2021-04-30
Add optimization result collection to basic result collection

SimonBoothroyd:
## Description
This PR adds a convenience method to create a basic data set collection from an existing optimization result collection. This would allow, for example, being able to find all of the hessian result records which were generated using the final structure of a set of optimizations.

This may be preferable in certain cases to just creating the basic result collection itself. This is especially true when trying to create a result collection from an older dataset which does not have CMILES information attached. This method will propagate the parent optimization CMILES down to the child basic result record thus circumventing this issue.

## Status
- [x] Ready to go

### DIFF
--- a/openff/qcsubmit/results/caching.py
+++ b/openff/qcsubmit/results/caching.py
@@ -45,7 +45,7 @@ def clear_results_caches():
     _grid_id_cache.clear()
 
 
-def _batched_indices(indices: List[T], batch_size: int) -> List[List[T]]:
+def batched_indices(indices: List[T], batch_size: int) -> List[List[T]]:
     """Split a list of indices into batches.
 
     Args:
@@ -59,7 +59,7 @@ def _batched_indices(indices: List[T], batch_size: int) -> List[List[T]]:
 
 
 @lru_cache()
-def _cached_fractal_client(address: str) -> FractalClient:
+def cached_fractal_client(address: str) -> FractalClient:
     """Returns a cached copy of a fractal client."""
     return FractalClient(address)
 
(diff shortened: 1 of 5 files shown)


jthorton (commented):
Thanks for taking this on, I think this is quite clean and very useful. My one concern is that I don't think we are checking the qcspec/keywords match those used for the optimisation dataset. You can include this in the query see [here](https://github.com/MolSSI/QCFractal/blob/5f00dd06bb34ca912c4055f0cbac60863ea89c7f/qcfractal/interface/client.py#L651) which saves us doing the filtering. I think its safe to assume users have used the same program, method, basis and keywords for both calculations so we should filter by these as well.

openff/qcsubmit/results/caching.py
```diff
@@ -102,11 +102,11 @@ def _cached_client_query(
         if (client_address, query_id) in query_cache
     ]
 
-    client = _cached_fractal_client(client_address)
+    client = cached_fractal_client(client_address)
 
     logger.debug(f"starting {query_name} to {client_address}")
 
-    batch_query_ids = _batched_indices(missing_query_ids, client.query_limit)
+    batch_query_ids = batched_indices(missing_query_ids, client.query_limit)
```
jthorton, line 109: Just wondering what the names have changed?

SimonBoothroyd: I'm starting to access these a bit outside of the `caching` module so I thought it was probably ok to just make them public members. I'd be happy to make them private though if that's your preference?

SimonBoothroyd:
> My one concern is that I don't think we are checking the qcspec/keywords match those used for the optimisation dataset. You can include this in the query see here which saves us doing the filtering. I think its safe to assume users have used the same program, method, basis and keywords for both calculations so we should filter by these as well.

Ah brilliant catch - thanks! I've (I think) added this in 02db5a6

jthorton (approved):
Thanks for adding the specification checks. LGTM!

Issue #42, opened 2020-09-25
Component Private settings.


jthorton:
In the multiprocessing update by @trevorgokey we introduce new per component parameters that should be immutable by users, mainly this is a flag which determines if the component produces duplicates and if we should run deduplication on the output of the component. This flag should be set in the constructor and should be hidden from the user as we don't want them to change it by accident. My idea is to use introduce a private `_settings` parameter to the base component which takes the following immutable pydantic class which can easily be extended to capture more private settings. While the `_settings` object can still be changed at runtime the effort required is much higher as the user has to import the class make a new settings instance and then swap it out, this way we ensure it is intentional if someone does this rather than accidentally turning off the deduplication stage while trying to set a different parameter.  This way the settings are also not exported to json which means someone can not export a workflow to json and change the file before loading it back in.

Are there any possible downsides to this approach? 

```python
class Settings(BaseModel):
    process_parallel: bool = True
    produces_duplicates: bool = True

    class Config:
        allow_mutation: bool = False
         extra: "forbid"
```

trevorgokey:
Yeah I think having a structure to keep the private bits separate would be good, especially if the idea is to omit them from serialization. We may opt to serialize a reference to a frozen set of settings, that can be pulled in/created automatically upon deserialization. I can see this being very useful for dataset versioning and keeping track of provenance for the workflows.

Would these be better served as calling them properties? Settings seems like data which should be user mutable. Properties seems like a more appropriate name in this context, and I think we can design them such that it is only possible to change the entire properties object, essentially making it a different component.

jthorton:
Yeah very happy to change the name and I think the above can only be modified by swapping out the entire settings object note the config settings. Ill add this in after we get your PR in as I am planning a slight refactor and improvement of the workflow components by moving to a more factory-like setup so users can inject new components easier for development and testing.

jthorton:
Now that pydantic supports private attributes update to use proper type hints for class vs private attributes.

PR #133, opened 2021-05-06
Component/Factory update

jthorton:
## Description
This PR updates the `wokflow_components` to use the new pydantic private attributes, we also reduce the amount of information serialised in the models by making any constant fields classmethods such as the description and the properties which decided if the component can be used in parallel. Now the component name is a literal string that distinguishes the component in the workflow and so the workflow is now a list of components that are executed in order. The factory dataset creation logic has also been simplified to remove duplicated code. 

Impliments #42 

## Status
- [x] Ready to go

### DIFF
--- a/openff/qcsubmit/factories.py
+++ b/openff/qcsubmit/factories.py
@@ -1,6 +1,6 @@
 import abc
 import os
-from typing import Any, Dict, List, Optional, Tuple, Type, TypeVar, Union
+from typing import Dict, List, Optional, Tuple, Type, TypeVar, Union
 
 import tqdm
 from openff.toolkit import topology as off
@@ -26,9 +26,12 @@ from openff.qcsubmit.exceptions import (
 )
 from openff.qcsubmit.procedures import GeometricProcedure
 from openff.qcsubmit.serializers import deserialize, serialize
-from openff.qcsubmit.workflow_components import CustomWorkflowComponent, get_component
+from openff.qcsubmit.workflow_components import (
+    Components,
+    CustomWorkflowComponent,
+    get_component,
+)
 
-C = TypeVar("C", bound=CustomWorkflowComponent)
 T = TypeVar("T", bound=BasicDataset)
 
 
@@ -37,14 +40,20 @@ class BaseDatasetFactory(CommonBase, abc.ABC):
     The Base factory which all other dataset factories should inherit from.
     """
 
-    factory_type: Literal["BaseDatasetFactory"] = Field(
+    type: Literal["BaseDatasetFactory"] = Field(
         "BaseDatasetFactory",
         description="The type of dataset factory which corresponds to the dataset made.",
     )
-    workflow: List[C] = Field(
+    workflow: List[Components] = Field(
         [],
         description="The set of workflow components and their settings which will be executed in order on the input molecules to make the dataset.",
     )
+
+    @classmethod
+    def from_file(cls, file_name: str):
+        """Create a factory from the serialised model file."""
+        data = deserialize(file_name=file_name)
+        return cls(**data)
 
     @classmethod
     @abc.abstractmethod
@@ -65,10 +74,8 @@ class BaseDatasetFactory(CommonBase, abc.ABC):
                 molecules=[
                     molecule,
                 ],
-                component_name="MolecularComplexRemoval",
-                component_description={
-                    "component_description": "Remove any molecules which are complexes."
-                },
+                component="MolecularComplexRemoval",
+                component_settings={},
                 component_provenance=self.provenance(),
             )
 
@@ -106,13 +113,7 @@ class BaseDatasetFactory(CommonBase, abc.ABC):
         """
         self.workflow = []
 
-    def add_workflow_component(
-        self,
-        components: Union[
-            List[C],
-            C,
-        ],
-    ) -> None:
+    def add_workflow_components(self, *components: Components) -> None:
         """
         Take the workflow components validate them then insert them into the workflow.
 
@@ -125,10 +126,6 @@ class BaseDatasetFactory(CommonBase, abc.ABC):
 
         """
 
-        if not isinstance(components, list):
-            # we have one component make it into a list
-            components = [components]
-
         for component in components:
             if issubclass(type(component), CustomWorkflowComponent):
                 try:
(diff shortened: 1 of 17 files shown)


openff/qcsubmit/tests/test_factories.py
```diff
@@ -91,7 +91,7 @@ def test_adding_workflow_components(factory_type):
     with pytest.raises(ValidationError):
         factory.workflow = {"first component": 3}
 
-    factory.workflow = {"test_conformer": conformer_gen}
+    factory.workflow = [conformer_gen, ]
```
SimonBoothroyd, line 94:
```suggestion
    factory.workflow = [conformer_gen]
```

openff/qcsubmit/tests/test_factories.py
```diff
@@ -169,7 +169,7 @@ def test_get_wrokflow_component(factory_type):
     factory.add_workflow_component(components)
 
     for component in components:
-        assert factory.get_workflow_component(component.component_name) == component
+        assert factory.get_workflow_component(component.component_name) == [component, ]
```
SimonBoothroyd, line 172:
```suggestion
        assert factory.get_workflow_component(component.component_name) == [component]
```

openff/qcsubmit/factories.py
```diff
@@ -28,41 +28,49 @@
...
 
-    factory_type: Literal["BasicDatasetFactory"] = Field(
-        "BasicDatasetFactory",
+    factory_type: Literal["BaseDatasetFactory"] = Field(
+        "BaseDatasetFactory",
         description="The type of dataset factory which corresponds to the dataset made.",
     )
-    workflow: Dict[str, CustomWorkflowComponent] = Field(
-        {},
+    workflow: List[C] = Field(
```
SimonBoothroyd, line 44: I think for this to work fully in pydantic this needs to be `List[Union[StandardConformerGenerator, MolecularWeightFilter, ..., etc]]` where all possible components are listed here

openff/qcsubmit/factories.py
```diff
@@ -125,55 +133,49 @@ def add_workflow_component(
...
-                    self.workflow[component.component_name] = component
 
             else:
                 raise InvalidWorkflowComponentError(
                     f"Component {component} rejected as it is not a sub "
                     f"class of CustomWorkflowComponent."
                 )
 
-    def get_workflow_component(self, component_name: str) -> CustomWorkflowComponent:
+    def get_workflow_component(self, component_name: str) -> List[C]:
```
SimonBoothroyd, line 149: Hmm maybe this should now be `get_workflow_components`?

openff/qcsubmit/factories.py
```diff
@@ -96,13 +104,13 @@ def clear_workflow(self) -> None:
         """
         Reset the workflow to by empty.
         """
-        self.workflow = {}
+        self.workflow = []
 
     def add_workflow_component(
```
SimonBoothroyd, line 109: Would it maybe make sense to change this to `add_workflow_components` a maybe allow it to accept an args list of components rather than just a list? i.e. `add_workflow_components(*components: Component)` that way you could write

```
factory.add_workflow_components(component_1, component_2, ...)
```

without needing to wrap them in a list?

openff/qcsubmit/factories.py
```diff
@@ -691,186 +587,118 @@ class TorsiondriveDatasetFactory(OptimizationDatasetFactory):
...
+        except KeyError:
+            dataset.filter_molecules(
+                molecules=[
+                    molecule,
+                ],
+                component_name="LinearTorsionRemoval",
+                component_description={
+                    "component_description": "Remove any molecules with a linear torsions selected to drive."
+                },
+                component_provenance=self.provenance(),
```
SimonBoothroyd, line 617: Do these need changing now (e.g. remove `component_description`)?

SimonBoothroyd (commented):
I think in general these changes look great! I've made a couple of suggestions (mainly based on personal preference) that you might consider.

openff/qcsubmit/factories.py
```diff
@@ -182,10 +184,11 @@ def remove_workflow_component(self, component_name: str) -> None:
             MissingWorkflowComponentError: If the component could not be found by its component name in the workflow.
         """
 
-        try:
-            del self.workflow[component_name]
-
-        except KeyError:
+        components = self.get_workflow_component(component_name=component_name)
```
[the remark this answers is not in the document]
jthorton: Nice catch thanks!

openff/qcsubmit/factories.py
```diff
@@ -28,41 +28,49 @@
...
+class BaseDatasetFactory(CommonBase, abc.ABC):
+    """
+    The Base factory which all other dataset factories should inherit from.
     """
 
-    factory_type: Literal["BasicDatasetFactory"] = Field(
-        "BasicDatasetFactory",
+    factory_type: Literal["BaseDatasetFactory"] = Field(
+        "BaseDatasetFactory",
         description="The type of dataset factory which corresponds to the dataset made.",
```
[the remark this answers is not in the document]
jthorton: Good idea, I agree the factory part is implied!

openff/qcsubmit/factories.py
jthorton: Yeah that's a good point and makes more sense with the return type now.

openff/qcsubmit/workflow_components/base_component.py
```diff
@@ -1,97 +1,61 @@
...
 
-    # this is a pydantic workaround to add private variables taken from
-    # https://github.com/samuelcolvin/pydantic/issues/655
-    __slots__ = [
-        "_cache",
-    ]
+    component_name: Literal["CustomWorkflowComponent"] = Field(
+        "CustomWorkflowComponent",
+        description="The name of the component which should match the class name.",
+    )
```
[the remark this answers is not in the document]
jthorton: Nice I think this looks much better!

openff/qcsubmit/factories.py
jthorton: Ah yeah just tested it and it does need to be a full list of components thanks for catching this.

openff/qcsubmit/factories.py
jthorton: Great idea adding this in.

openff/qcsubmit/factories.py
jthorton: Good point, yes I think these should be removed to be consistent with the new components.

PR #138, opened 2021-05-20
LowestEnergyFilter

jthorton:
## Description
This PR adds the `LowestEnergyFilter` which can be used on Basic and Optimisation results collections to reduce down the number of repeated entries for a molecule to a single entry corresponding to the lowest energy structure. 

## Todos
  - [x] Add tests for this filter

## Status
- [x] Ready to go

### DIFF
--- a/openff/qcsubmit/results/filters.py
+++ b/openff/qcsubmit/results/filters.py
@@ -1,17 +1,26 @@
 import abc
 import logging
 from collections import defaultdict
-from typing import List, Optional, Set, TypeVar, Union
+from typing import List, Optional, Set, Tuple, TypeVar, Union
 
 import numpy
 from openff.toolkit.topology import Molecule
 from pydantic import BaseModel, Field, PrivateAttr, root_validator, validator
 from qcelemental.molutil import guess_connectivity
-from qcportal.models.records import RecordBase, RecordStatusEnum
+from qcportal.models.records import (
+    OptimizationRecord,
+    RecordBase,
+    RecordStatusEnum,
+    ResultRecord,
+)
 from simtk import unit
 from typing_extensions import Literal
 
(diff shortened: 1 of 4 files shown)


openff/qcsubmit/results/filters.py
```diff
@@ -148,6 +157,75 @@ def _apply(self, result_collection: "T") -> "T":
         return result_collection
 
 
+class ResultRecordGroupFilter(ResultFilter, abc.ABC):
+    """The base class for filters which reduces repeated molecule entries down to a single entry."""
```
SimonBoothroyd, line 161:
```suggestion
    """The base class for filters which reduces repeated molecule entries down to a single entry.
    
    Notes:
        * This filter will only be applied to basic and optimization datasets. Torsion drive datasets / entries will be skipped.
    """
```

openff/qcsubmit/results/filters.py
```diff
@@ -148,6 +157,75 @@ def _apply(self, result_collection: "T") -> "T":
...
+            )
+            filtered_results[address].append(result)
+
+        result_collection.entries = filtered_results
+
+        return result_collection
+
+
+class LowestEnergyFilter(ResultRecordGroupFilter):
+    """Filter the results collection and only keep the lowest energy entries."""
```
SimonBoothroyd, line 206:
```suggestion
    """Filter the results collection and only keep the lowest energy entries.
    
    Notes:        
        * This filter will only be applied to basic and optimization datasets. Torsion drive datasets / entries will be skipped.
```

SimonBoothroyd (approved):
Thanks for adding this in - I think it will come in handy a lot! LGTM

openff/qcsubmit/results/filters.py
jthorton: Good catch!

PR #139, opened 2021-05-25
Add conformer RMSD result filter

SimonBoothroyd:
## Description

This PR adds a new filter which aims to retain up to a maximum number of maximally distinct conformers for each molecule in an optimization or a basic result collection while conformers similar to within an RMSD threshold are discarded.

## Status
- [x] Ready to go

### DIFF
--- a/openff/qcsubmit/results/filters.py
+++ b/openff/qcsubmit/results/filters.py
@@ -364,7 +364,10 @@ class ConformerRMSDFilter(ResultRecordGroupFilter):
         assert all(molecule.n_conformers == 1 for _, _, molecule, _ in entries)
 
         # Condense the conformers into a single molecule.
-        conformers = [molecule.conformers[0] for _, _, molecule, _ in entries]
+        conformers = [
+            molecule.canonical_order_atoms().conformers[0]
+            for _, _, molecule, _ in entries
+        ]
 
         [_, _, molecule, _] = entries[0]
 
--- a/openff/qcsubmit/results/filters.py
+++ b/openff/qcsubmit/results/filters.py
@@ -1,4 +1,6 @@
 import abc
+import copy
+import itertools
 import logging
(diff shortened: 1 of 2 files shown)


openff/qcsubmit/results/filters.py
```diff
@@ -158,19 +160,20 @@ def _apply(self, result_collection: "T") -> "T":
...
+          Torsion drive datasets / entries will be skipped.
     """
 
     @abc.abstractmethod
     def _filter_function(
         self, entries: List[Tuple["_BaseResult", RecordBase, Molecule, str]]
-    ) -> Tuple["_BaseResult", str]:
-        """A method which should reduce a set of results down to a single entry based on some property of the QC
-        calculation.
+    ) -> List[Tuple["_BaseResult", str]]:
```
jthorton, line 174: Ahh good catch!

openff/qcsubmit/results/filters.py
```diff
@@ -194,13 +197,16 @@ def _apply(self, result_collection: "T") -> "T":
         filtered_results = defaultdict(list)
 
         for entries in entries_by_inchikey.values():
-            result, address = self._filter_function(
+
+            results_and_addresses = self._filter_function(
```
jthorton, line 201: Nice!

jthorton (commented):
Thanks for adding this I think this will be very useful. One thing I am not sure about is, what happens if a molecule is in two different datasets in a different atom ordering when trying to compare the RMSD? Should the molecule given by the `to_records` function put all of the molecules into canonical order depending on what toolkit is being used to do the filtering?

SimonBoothroyd:
Thanks @jthorton for the review!

> Should the molecule given by the to_records function put all of the molecules into canonical order depending on what toolkit is being used to do the filtering?

Great catch! I've fixed this in fd61bf9

PR #141, opened 2021-06-17
Add API to summarize result collection SMIRNOFF coverage

SimonBoothroyd:
## Description

This PR adds a new `smirnoff_coverage` API point to the result collection objects which returns a summary of how many unique molecules exercise each parameter within a SMIRNOFF force field.

Additionally for torsion drive result collections, an option is exposed to only return coverage for parameters which are at least partially applied across the driven central bond.

## Known Issues

The code which reports coverage across driven torsions does not currently deduplicate torsions performed for the same molecule but which have a different ordering.

## Status
- [X] Ready to go


jthorton:
Good idea to add this to the results class as well, I wonder if it is worth making this a utility function that can be shared between the dataset class and results class as this is currently implemented [here](https://github.com/openforcefield/openff-qcsubmit/blob/fe42e13451610a51c86b0755e0d587d9b5be2dff/openff/qcsubmit/datasets/datasets.py#L915) with slightly different functionality. It might be nice to combine the two and offer users the ability to provide a list of forcefields as well as the option to only look at terms through torsions.

SimonBoothroyd:
> Good idea to add this to the results class as well, I wonder if it is worth making this a utility function that can be shared between the dataset class and results class as this is currently implemented here with slightly different functionality. It might be nice to combine the two and offer users the ability to provide a list of forcefields as well as the option to only look at terms through torsions.

Great idea - I've split this functionality into a new utility module and now the data set and result functions just reference this.

jthorton (approved):

PR #147, opened 2021-06-24
Make `long_description_url` metadata optional

SimonBoothroyd:
## Description

This PR makes the `Metadata.long_description_url` optional as it is likely unneeded for private and local submissions.

## Status
- [X] Ready to go


jthorton (approved):
Good idea I think this helps avoid a common pattern of setting the URL to something like `https://test.org` which I was doing in a lot of tests that are now failing. Feel free to merge once fixed.

PR #148, opened 2021-06-25
Fix retrieving data from snowflake server

SimonBoothroyd:
## Description

This PR fixes a bug where data could not be retrieved from a local QCFractal snowflake due to the client verification failing and hence the connection being refused.

## Status
- [X] Ready to go


jthorton (approved):
Ahh great catch this will help a lot with the bespoke work and local instances.

Issue #59, opened 2020-11-05
Torsiondrive Factory Torsion Selection


jthorton:
While some modules like the Coverage filter and the SMARTS filter allow users to declare which types of torsions are allowed to be driven in torsiondrive datasets this is too specific for exploratory datasets where we wish to rotate all torsions. By default, this case is covered by the torsiondrive factory but it also selects trivial rotations like methyl groups. I think it would be best to add a setting to the factory to allow users to specify certain groups they do not wish to rotate this could be an enum of predefined types or groups of types or we could allow smarts patterns or both. 

This is not a problem when we use fragmenter to make a torsiondrive dataset as it excludes certain trivial terminal torsions.

jthorton:
I think it might be best to have an extra argument in the torsiondrive factory which can take custom or predefined smarts patterns for certain groups we do not wish to be driven something like this

```python
from qcsubmit.factories import TorsiondriveDatasetFactory
from qcsubmit import FunctionalGroups as fg

factory = TorsiondriveDatasetFactory()

dataset = factory.create_dataset(ignore_torsions=[fg.Methyl, "my_smarts", ...])
```
The factory will then remove any entries from the dataset whose target torsion central bond matches with any of the supplied smarts patterns. If not an argument in the function this could be an attribute on the factory class maybe, this would also allow for the validation of the smarts before running the workflow, ie checking the number of tagged atoms is either 2 or 4.

jchodera:
This sounds great!

Would the specified SMARTS be _tagged_ SMARTS delineating the torsion, such as `[*:1]~[*:2]-[CH3:3]-[H:4]` for a methyl torsion?

jthorton:
Unfortunately, this still needs to be fixed but is now further complicated by the fact we want to support 2D scans involving any combination of two degrees of freedom. I think I am now moving to the idea of two new workflow components which can handle 1D and 2D scans respectively. These components should be able to handle multiple allowed smarts with their own distinct scan ranges I imagine something like this in the 1D case

```python
from openff.qcsubmit.workflow_components import ScanFilter1D

filter = ScanFilter1D()
# add a methyl full scan
filter.add_scan(smarts="[*:1]~[*:2]-[CH3:3]-[H:4]", range=(-165, 180))
# add a limited ring scan
filter.add_scan(smarts="[*:1]~[*:2]:[*:3]~[*:4]", range=(-40, 40)) 
# inspect the first scan
filter.scans[0]
``` 
[...]

PR #149, opened 2021-06-25
Torsion selection rework

jthorton:
## Description
This PR implements the `ScanFilter` and `ScanEnumerator` proposed in #59 by @SimonBoothroyd as a way to tag DOF for scanning. This also changes the default behaviour of the `TorsiondriveDatasetFactory` to not select every rotatable bond in a molecule for scanning. The factory will now only accept molecules that have torsions tagged for scanning and will remove any molecules not tagged. Users should now use the `ScanEnumerator` to select torsions by putting it in a workflow.

## Notes 
- if the enumerator would match multiple times to a molecule only the first occurrence returned by `openff.toolkit.topology.Molecule.chemical_environment_matches` will be tagged for scanning. 
- If a molecule has multiple torsions tagged for scanning and matches the `ScanFilter` only the offending torsion will be removed meaning the molecule can still be put in a dataset if it has at least one valid torsion scan. `ScanFilter` will only remove the molecule from a dataset if the molecule has no torsions tagged after filtering. 

## Todos
  - [ ] TODO 1

## Questions
- [ ] Should we remove the ability to select torsions using the [SmartsFilter](https://github.com/openforcefield/openff-qcsubmit/blob/3816027646ad5b725ad6601f02c6e396eb35d1b9/openff/qcsubmit/workflow_components/filters.py#L427) and [Coveragefilter](https://github.com/openforcefield/openff-qcsubmit/blob/3816027646ad5b725ad6601f02c6e396eb35d1b9/openff/qcsubmit/workflow_components/filters.py#L237) to keep the logic in one place to select torsions (`ScanEnumerator`) ?

## Status
- [ ] Ready to go


openff/qcsubmit/workflow_components/base.py
```diff
@@ -145,21 +152,22 @@ def list_components() -> List[Components]:
 
 # register the current components
 # conformer generation
-register_component(StandardConformerGenerator())
```
SimonBoothroyd, line 148: 🎉

jthorton: Thought you might like this, thanks for showing me another way to do it in bespokefit!

SimonBoothroyd (approved):
Thanks for adding this @jthorton - I think this will be a big QoL improvement! Especially given the changes in b08f781. This LGTM!

PR #150, opened 2021-06-29
Expose keywords via the QCSpec

SimonBoothroyd:
## Description

This PR adds the ability to add optional `keywords` to datasets via the associated `QCSpec` specifications.

## Status
- [X] Ready to go


jthorton (approved):
LGTM, great job I wonder if we should also move the `maxiter` and `scp_properties` to the `QCSpec` class in another PR? Currently, they get applied to all specs on a dataset however for MM and ANI specs these keywords don't make sense.  They don't change the calculation however as they are just ignored.

PR #151, opened 2021-06-29
Charge filter

jthorton:
## Description
This PR adds a simple `FormalChargeFilter` which can filter molecules by net formal charges. The molecules are kept or removed based on the list of `allowed_charges` or `filtered_charges`. 

## Status
- [ ] Ready to go


SimonBoothroyd (approved):
I think this is a nice addition! 

The only thing you might consider is changing `FormalChargeFilter` -> `ChargeFilter`, `allowed_charges` -> `charges_to_include` and `filtered_charges` -> `charges_to_exclude` to be consistent with the results filter, but I don't think this is a big deal!

Otherwise LGTM!

PR #157, opened 2021-08-12
Dihedral selection update

jthorton:
## Description
This PR updates the torsiondrive deduplication method to also take into account the symmetry of the molecule. Now only one torsion per unique symmetry group is stored and scanned. The coverage and smarts filters have had their ability to tag torsions removed to simplify the classes as torsion selection should now be done using the scan filters. 

Example:
For a molecule that has multiple torsions highlighted which are not symmetry unique the current deduplication method will result in the following dataset with 8 torsiondrives 
image
After the deduplication update, we get the following dataset with only 6 torsiondrives as expected
image

## Todos
  - [x] simplify the coverage and smarts filters to only accept allowed or disallowed items not both in one instance
  - [x] add tests 

## Status
- [x] Ready to go


openff/qcsubmit/workflow_components/filters.py
```diff
@@ -205,15 +205,8 @@ def provenance(self) -> Dict:
 
 class CoverageFilter(BasicSettings, CustomWorkflowComponent):
     """
-    Filters molecules based on the requested forcefield coverage.
+    Filters molecules based on the requested force field parameter ids..
```
SimonBoothroyd:
```suggestion
    Filters molecules based on the requested force field parameter ids.
```

SimonBoothroyd (approved):
This is an awesome QoL update - thanks for adding this! LGTM!

PR #158, opened 2021-08-23
Toolkit registry

jthorton:
## Description
This PR allows the workflow components and the dataset factories to take a toolkit registry object that declares the available backend toolkits and their querying presence for functionality. If `None` is passed the default global toolkit registry will be used which contains all currently installed toolkits. This replaces the per-component toolkit argument. The provenance information now contains all toolkits in the passed toolkit registry. 

## Status
- [x] Ready to go


SimonBoothroyd (approved):
Great idea - LGTM!

PR #160, opened 2021-09-02
Update OpenMM imports for 7.6 and backwards compatibility

mattwthompson:
I noticed tests were failing due to the API breaks in OpenMM 7.6. I tried to update everything to work with the new version (avoiding deprecation warnings) but also be compatible with 7.5.x. I was unable to test everything locally but the major one in `ElementFilter` should be fixed.


jthorton (approved):
@mattwthompson Thanks for catching all of those it looks like we are just waiting on a new QCEngine release now to fix the other testing errors, feel free to merge.

mattwthompson:
Sounds good - thanks for the quick review @jthorton!

PR #162, opened 2021-09-07
Add minimum N conformers result filter

SimonBoothroyd:
## Description

This PR adds a new filter that only retains records for associated with a particular unique molecule if that molecule has a certain number of records (i.e. conformers) associated with it.

This is useful, for example, when you want to train to relative energies and hence need to make sure there are at least 2 conformers present. This is especially true when used in conjunction with the `ConformerRMSDFilter`.

## Status
- [ ] Ready to go


jthorton (approved):
LGTM another great addition!

PR #165, opened 2021-09-20
HDF5 support

jthorton:
## Description
This PR adds support for HDF5 files with a specific format which will be explained in the documentation. We also add basic support for multi molecule inputs.

## Todos
  - [x] add tests for hdf5
  - [x] add docs explaining the format

## Status
- [x] Ready to go


jthorton:
The broken psi4 tests seem to coincide with a qcelemental release which I will look into in another PR.

PR #173, opened 2021-10-14
Added allowance for list of floats to QCSpec.keywords

dotsdl:
## Description
In openforcefield/qca-dataset-submission#230, we encountered a case where a list of floats must be given for a keyword, namely `PERTURB_DIPOLE`. This change allows for a list of floats as a valid `QCSpec.keyword` value.

Are there any other places in the codebase this change should be propagated to? I locally tested out [the submission notebook](https://github.com/openforcefield/qca-dataset-submission/blob/d0718496e563e02e1f2902cfacc122ef58564bc1/submissions/2021-10-01-OpenFF-resppol-mp2-single-point/generate-dataset.ipynb) with this change, avoiding giving a string as the value for `PERTURB_DIPOLE`, but haven't performed tests beyond this.

## Status
- [x] Ready to go

### DIFF
--- a/openff/qcsubmit/tests/test_datasets.py
+++ b/openff/qcsubmit/tests/test_datasets.py
@@ -15,7 +15,7 @@ try:
 except ImportError:
     from simtk import unit
 
-from openff.qcsubmit.common_structures import MoleculeAttributes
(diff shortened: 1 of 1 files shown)


dotsdl:
Awesome, thank you so much @jthorton!

PR #183, opened 2021-12-08
Only call `Molecule.from_openff_molecule` if 'attributes' not present in `Dataset.add_molecule` kwargs

dotsdl:
Currently, [we observe failures](https://github.com/openforcefield/qca-dataset-submission/runs/4459934379?check_suite_focus=true#step:8:24) when calling `Dataset.add_molecule` when feeding in `molecule=None`, whereas this was permitted previously.

This appears due to the use of `kwargs.pop`, with `MoleculeAttributes.from_openff_molecule(molecule=molecule)` as the default value. This gets called even if 'attributes' is present as a key in the dictionary. This PR separates these operations to avoid this call.

## Status
- [x] Ready to go

### DIFF
--- a/openff/qcsubmit/datasets/datasets.py
+++ b/openff/qcsubmit/datasets/datasets.py
@@ -595,7 +595,7 @@ class _BaseDataset(abc.ABC, CommonBase):
     def add_molecule(
         self,
         index: str,
-        molecule: off.Molecule,
+        molecule: Optional[off.Molecule],
         extras: Optional[Dict[str, Any]] = None,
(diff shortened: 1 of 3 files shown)


dotsdl:
@jthorton this is currently blocking `qca-dataset-submission` validation. I will temporarily change validation there to use this branch instead.

jthorton:
@dotsdl Thanks for fixing this I did not realise that the default statements were evaluated as well Ill add a test to ensure I don't break this again and then merge!

PR #186, opened 2022-02-02
Removed chunked saves for entry submission; added performance hack for single point task submission

dotsdl:
## Description
This removes chunked calls to `dataset.save` as entries (molecules) are added to a given dataset as part of submission, as these are unnecessary and only serve to slow submission as the collection object grows with additional entries added.

We also added in a performance hack for single point `Dataset`s only, in which we only call `collection.save` on the first call for a given spec addition and then again after all tasks are generated. This saves repeated movement of the collection object, which is unchanged as tasks are generated over the course of the loop.

Requires a new release of QCFractal beyond the current 0.15.7 release, in particular MolSSI/QCFractal#705.

## Status
- [ ] Ready to go

### DIFF
--- a/openff/qcsubmit/common_structures.py
+++ b/openff/qcsubmit/common_structures.py
@@ -339,7 +339,7 @@ class PCMSettings(ResultsConfig):
             if "cavity_Area" not in kwargs:
                 cavity_Area = (
                     self.__fields__["cavity_Area"].default
-                    * constants.bohr2angstroms ** 2
+                    * constants.bohr2angstroms**2
                 )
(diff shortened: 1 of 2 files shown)


jthorton:
@dotsdl thanks for adding this speed up! Looks like we need a new version of qcfractal to test on do we know when we will have a realese?

dotsdl:
@jthorton hopefully today!

jthorton:
@dotsdl merged the other PR feel free to merge this one once tests pass

dotsdl:
Thanks @jthorton! Merging!

PR #192, opened 2022-03-08
keep multiple conformers when creating a basic dataset from an `OptimizationResultCollection`

jthorton:
## Description
This PR fixes a bug where an `OptimizationResultCollection` will drop multiple conformers for molecules when creating a basic dataset from the result. 

## Status
- [x] Ready to go


SimonBoothroyd (approved):
Looks great thanks for fixing this!

PR #194, opened 2022-03-09
Expose trajectory protocol for optimization datasets

dotsdl:
## Description

We set the default trajectory protocol for optimizations to be
`initial_and_final` in order to conserve space where possible.
Choosing another option is an opt-in choice for the submitter.

## Status
- [ ] Ready to go

### DIFF
--- a/openff/qcsubmit/common_structures.py
+++ b/openff/qcsubmit/common_structures.py
@@ -26,6 +26,7 @@ from pydantic import (
 from qcelemental import constants
 from qcelemental.models.common_models import Model
 from qcelemental.models.results import WavefunctionProtocolEnum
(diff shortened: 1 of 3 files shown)


dotsdl:
Going to wait on this for the QCFractal refactor, as there are literally [two](https://github.com/MolSSI/QCFractal/blob/ceff64fe032590095e0f865bc1d0c2da4684404e/qcfractal/interface/collections/optimization_dataset.py#L32) [places](https://github.com/MolSSI/QCFractal/blob/ceff64fe032590095e0f865bc1d0c2da4684404e/qcfractal/interface/models/common_models.py#L271) where protocols should go on an optimization entry, and it is not clear right now which one to use. We could try just attaching to both, but this may prove to be a temporary solution anyway until the refactor is out.

PR #196, opened 2022-04-14
Export basic datasets to hdf5

jthorton:
## Description

This PR adds the ability to export basic datasets to hdf5 files, currently, we extract the energies and gradients but more properties could be included in future.

### Layout 

Each molecule has its own group under its fixed hydrogen layer inchikey with the following datasets:
  - `smiles`: The mapped explicit hydrogen smiles which can be used to construct the molecule via the openff-toolkit using `Molecule.from_mapped_smiles`
  - `atomic_numbers`: An array of the atomic numbers type int16
  - `charge`: The total charge on the molecule calculated by the openff-toolkit as the sum of formal charges, type int16
  - `specification`: The `method:basis` used to compute the results type h5py string.
  - `energies`: An array of energies for the molecule in units of `hartree` in the same order as the conformations, so `energies[i]` corresponds to `conformation[i]` type float64
  - `conformations`: An array of conformations for the molecule in `bohr` type float64
  - `gradients`: An array of gradients in units `hartree / bohr` in the same order as the conformations, so `gradients[i]` corresponds to `conformation[i]` type float64.

## Todos
  - [ ] add tests

## Questions
- [ ] should we extract any other properties?

## Status
- [ ] Ready to go

### DIFF
--- a/openff/qcsubmit/results/results.py
+++ b/openff/qcsubmit/results/results.py
@@ -407,6 +407,91 @@ class BasicResultCollection(_BaseResultCollection):
 
         return records_and_molecules
 
+    def to_hdf5(self, filename: str):
+        """
+        Write the dataset to a custom HDF5 file format suitable for ML purposes, the format is compact where
+        possible to avoid repeated redundant information.
+        """
+        import h5py
+        import numpy as np
+        import tqdm
+        from openmm import unit
+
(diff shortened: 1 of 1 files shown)


mattwthompson:
I know this is old, but I bet this is still doable with a quick rebase/update. Still interested?

Issue #201, opened 2022-08-25
`cached_query_torsion_drive_results` fails for multi-dimensional grid ids


chapincavender:
If a `TorsionDriveResultCollection` contains TorsionDrives with multi-dimensional grid indices, `cached_query_torsion_drive_results` will fail [here](https://github.com/openforcefield/openff-qcsubmit/blob/e4d5add8f169b7c358fb518358f12699923cee4d/openff/qcsubmit/results/caching.py#L404) with `ValueError: could not convert string to float`.

The TorsionDrive grid indices are stored as stringified tuples such as `'(-165)'`. For multi-dimensional TorsionDrives, the grid indices look like `'(-165, -165)'` so that `float(x[1:-1])` fails. I think something like
```        grid_ids.sort(key=lambda s: tuple(float(x) for x in s[1:-1].split(", ")))```
will work for both one-dimensional and multi-dimensional grid indices.

I have this patch in a new branch that I'm testing now. Any ideas for a solution that handles this problem more elegantly?

**Comment by dotsdl** [comment]
Thanks @chapincavender! Can you make a PR with your change? I think your approach is close to what I might do, but I'd probably go with:
```
grid_ids.sort(key=lambda s: tuple(float(x) for x in s.strip('()').split(',')))
```
This avoids any assumptions on slicing.

chapincavender:
Fixed in #202. It looks like the grid indices are stringified lists instead of stringified tuples as I stated above, so I went with `split("[]")` instead of splicing.

mattwthompson:
Closed via #202 and released in version 0.3.2

PR #202, opened 2022-08-29
Support multi-dimensional TorsionDrive grid indices

chapincavender:
## Description
This PR addresses #201 by supporting TorsionDrives with multi-dimensional grid indices.

## Status
- [ ] Ready to go

### DIFF
--- a/openff/qcsubmit/results/caching.py
+++ b/openff/qcsubmit/results/caching.py
@@ -401,7 +401,7 @@ def cached_query_torsion_drive_results(
 
         grid_ids = [*qc_record.minimum_positions]
         # order the ids so the conformers follow the torsiondrive scan range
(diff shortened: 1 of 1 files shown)


jthorton:
@chapincavender Thanks for taking this on it looks great! I think we just need a test which involves a 2d torsion drive and this will be good to merge!

chapincavender:
Thanks @jthorton! Do you think I can base the test on [this one](https://github.com/openforcefield/openff-qcsubmit/blob/e4d5add8f169b7c358fb518358f12699923cee4d/openff/qcsubmit/tests/results/test_caching.py#L142) and pull down one of my 2-D amino acid datasets?

dotsdl:
This fixes an issue with string slicing; it works fine for the more common 1-D case, so I'm fine with merging as is.

PR #206, opened 2022-11-08
Improve compatibility with non-Psi4 engines

mattwthompson:
## Description
The current logic makes it difficult to use QC packages supported by QCEngine besides Psi4. This updates the logic to _not_ assume that the `program` argument is in the `settings` dictionary.

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/devtools/conda-envs/docs.yaml
+++ b/devtools/conda-envs/docs.yaml
@@ -24,7 +24,7 @@ dependencies:
 #  - rdkit                MOCKED
   - pydantic
   - pyyaml
-  - qcportal
+  - qcportal >=0.15
   - torsiondrive
   - basis_set_exchange
   - typing-extensions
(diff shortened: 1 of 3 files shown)


jthorton (approved):
This should work great for everything else!

mattwthompson:
The docs builds were broken because an old QCPortal was being brought in, somehow. 6553fc4183b21ac63bdb2af733eefa7a33587cfd should fix that, but there might be other issues.

mattwthompson:
The change in 7e4d0911fb1dbc0a065bf4759821c0dd3e4a990f is a fairly naive way of getting tests to pass, but if I understand the change correctly it's in line with the intended changes; I think that previously the supported methods were implied by the keys in the `settings` and anything not recognized produced an error. The point of this change is to allow unrecognized methods through - presumably letting some errors pass through if users are using methods outside of the scope of current QCSubmit usage.

(I'm maybe 80% sure I'm using the right jargon here - method/program/settings/etc. - please correct me if I'm mistaken!)

jthorton:
> The point of this change is to allow unrecognized methods through

Almost but its programs, the test was checking if an error was raised if an unsupported program of QCEngine was supplied. It should have probably been checking with qcengine if the program is supported but this will also limit the programs to only those supported in the main version and not ones supplied via plugins, but that might be the best we can do for now.

PR #220, opened 2023-07-13
Use Psi4 1.8

mattwthompson:


jthorton:
Hey @mattwthompson thanks for looking into this, I think until Psi4 on conda-forge includes the fortran interface for PCM these tests will fail. We should keep an eye open for when this change happens but until then we need to use the psi4 channel.

mattwthompson:
You seem to be correct: https://psicode.org/psi4manual/1.5.0/glossary_psivariables.html#psivar-SCF-DIPOLE

mattwthompson:
The tests are running really, really slowly now: https://github.com/openforcefield/openff-qcsubmit/actions/runs/5554756531/jobs/10144963963

I haven't looked into why

j-wags:
Superseded by the already-merged #218

PR #226, opened 2023-07-25
Testing fix and update for Psi4 >1.5

jthorton:
## Todos
  - [x] Fix hanging test suite
  - [x] Update interface to work with new keywords in psi4 1.8

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/devtools/conda-envs/psi4.yaml
+++ b/devtools/conda-envs/psi4.yaml
@@ -1,15 +1,14 @@
 name: qcsubmit-test-psi4
 
 channels:
-  - psi4
-  - defaults
   - conda-forge
+  - conda-forge/label/libint_dev  # needed for psi4/libint2
   - openeye
 
 dependencies:
 
   # Base depends
-  - python=3.8
+  - python>=3.8
   - pip
 
   # Testing
@@ -25,15 +24,12 @@ dependencies:
 
   - openeye-toolkits
 
-  # Psi4 dependencies
-  - psi4 =1.5
-  - libint2=*=hc9558a2_9
-  - intel-openmp!=2019.5
-  - gau2grid >=2.0
+  - psi4 =1.8
 
   ### Core dependencies.
 
-  - openff-toolkit-base >= 0.11.3
+  - openff-toolkit-base == 0.14.0
+  - openff-units >=0.2.1
   - rdkit
   - pydantic =1
   - pyyaml
--- a/devtools/conda-envs/psi4.yaml
+++ b/devtools/conda-envs/psi4.yaml
@@ -28,7 +28,8 @@ dependencies:
 
   ### Core dependencies.
 
-  - openff-toolkit-base >= 0.11.3
+  - openff-toolkit-base >= 0.12
+  - openff-units >=0.2.1
   - rdkit
   - pydantic =1
   - pyyaml
--- a/devtools/conda-envs/psi4.yaml
+++ b/devtools/conda-envs/psi4.yaml
@@ -28,7 +28,7 @@ dependencies:
 
   ### Core dependencies.
 
-  - openff-toolkit-base >= 0.12
+  - openff-toolkit-base == 0.14.0
   - openff-units >=0.2.1
   - rdkit
   - pydantic =1
--- a/devtools/conda-envs/psi4.yaml
+++ b/devtools/conda-envs/psi4.yaml
@@ -1,14 +1,15 @@
 name: qcsubmit-test-psi4
 
 channels:
+  - psi4
+  - defaults
   - conda-forge
-  - conda-forge/label/libint_dev  # needed for psi4/libint2
   - openeye
 
 dependencies:
 
   # Base depends
-  - python>=3.8
+  - python=3.8
   - pip
 
   # Testing
@@ -24,12 +25,15 @@ dependencies:
 
   - openeye-toolkits
 
-  - psi4 =1.8
+  # Psi4 dependencies
+  - psi4 =1.5
+  - libint2=*=hc9558a2_9
+  - intel-openmp!=2019.5
+  - gau2grid >=2.0
 
   ### Core dependencies.
 
-  - openff-toolkit-base == 0.14.0
-  - openff-units >=0.2.1
+  - openff-toolkit-base >= 0.11.3
   - rdkit
   - pydantic =1
   - pyyaml
--- a/devtools/conda-envs/psi4.yaml
+++ b/devtools/conda-envs/psi4.yaml
@@ -8,7 +8,7 @@ channels:
 dependencies:
 
   # Base depends
-  - python =3.9
+  - python
   - pip
 
   # Testing
--- a/devtools/conda-envs/psi4.yaml
+++ b/devtools/conda-envs/psi4.yaml
@@ -8,7 +8,7 @@ channels:
 dependencies:
 
   # Base depends
-  - python>=3.8
+  - python =3.9
   - pip
 
   # Testing
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -69,7 +69,7 @@ jobs:
 
       - name: Run Tests
         run: |
-          pytest -v --cov=openff --cov-config=setup.cfg openff/qcsubmit/tests/ --cov-report=xml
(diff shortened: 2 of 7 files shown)


jthorton:
Tests seem to be hanging when finished going by the logs which report all tests as being run. It seems like the qcfractal testing server fails to shut down which stops the tests from finishing. I can reproduce this locally when running the full set of tests.

devtools/conda-envs/psi4.yaml
```diff
@@ -8,7 +8,7 @@ channels:
 dependencies:
 
   # Base depends
-  - python>=3.8
+  - python =3.9
```
jthorton, line 11: I don't think we need to pin the python version, this should be handled in the CI.yaml where we define the versions of python we want to test on?

ntBre: Hmm, that might be true, at least for the CI. When I had >=3.8 it failed locally because it installed 3.10 and we ran into the issue with `collections` I mentioned in the basic commit. Mine is failing locally anyway because of PCMSolver, so we can try it without the pin if you want.

ntBre: I guess we could do something like >=3.8, <3.10, but I think psi4 1.8 also requires >=3.9, so the end result should be 3.9 one way or another.

ntBre: I'm fine with unpinning for CI, but should we add additional conda-envs for local testing then? If we unpin the Python version, anyone trying to develop locally will get a broken environment.

Also if we unpin here, my earlier change for the basic tests needs to be unpinned as well, which is, again, fine with me. I just want to be able to set up a dev environment locally when working on this, and I think it would be best if those steps were encoded in the repo itself. Maybe that point is irrelevant once we can actually handle Python 3.10, though.

mattwthompson: What we've sorta settled on is having multiple environment YAML files for different environments that need significantly different sets of upstreams, i.e.

* https://github.com/openforcefield/openff-models/tree/main/devtools/conda-envs
* https://github.com/openforcefield/openff-forcefields/tree/main/devtools/conda-envs

and having the [workflow YAML](https://github.com/openforcefield/openff-forcefields/blob/e3b4387c3898412cdb5c68533ac81981bf7dd316/.github/workflows/CI.yml#L51-L52) be the single source of truth for Python versions in addition to constraints on upstream dependencies.

What's worked well for us is to assume individual developers' environments will diverge from what CI uses - I bring in a lot of tools that CI doesn't need, for example - but that everybody should start from there. After all, getting CI passing is a (soft) requirement for changes to be accepted. Sometimes I track a development environment for my own purposes, but this is neither something that CI sees nor something that I'd require other developers to start from.

I was able to solve the environment from this branch pretty quickly (`CONDA_SUBDIR=osx-64 mamba env create --file devtools/conda-envs/basic.yaml`) and get Python 3.9. I might be missing something, but is there an unmet need that would be satisfied by tracking a separate testing environment meant for local installs?

ntBre: > I was able to solve the environment from this branch pretty quickly (CONDA_SUBDIR=osx-64 mamba env create --file devtools/conda-envs/basic.yaml) and get Python 3.9

Yes, the basic environment is currently pinned on 3.9 too, so I would expect that you did get Python 3.9. If I take that basic.yaml and replace `python =3.9` with `python >=3.8` as proposed here and run `mamba env create -f devtools/conda-envs/basic.yaml`, I end up with Python 3.10.12, which means the tests will fail with the import error from collections.

> I might be missing something, but is there an unmet need that would be satisfied by tracking a separate testing environment meant for local installs?

Personally, I think it makes more sense *not* to have a separate local environment, but to ensure the CI environment is usable for local testing, which is the case with =3.9 and not the case without it. How else are you supposed to know if your changes pass the tests on CI without pushing them?

The separate environment suggestion was just an idea since it didn't seem like we wanted to make the CI environment usable locally.

I'm new to this and probably misunderstanding something. I just want a way to know that my local changes are going to pass CI before pushing them. I assumed that this meant I needed to be running in the same environment as the CI, but I am open to other options since y'all have obviously been managing this since before I joined.

More broadly, it feels to me that we should have something like fully-pinned environments that we know work. Nearly everything I've tried to do so far has required lots of environment debugging even to get code to run, much less to get tests to pass. Maybe these shouldn't be the normal CI environments, but I just feel like they should exist and be tracked somewhere with the code they run.

j-wags: I hadn't thought about this before. Given how useful it is to be able to use `mamba env create -f devtools/conda-envs/xxx.yaml` to make developer builds, I think it could be useful to BOTH:

* continue pinning exact versions of python in the CI build yamls (in the `.github` directory), AND
* specify a python version range in the yamls in `conda_envs` directory, for local installs

In this case, the python version range in the `conda_envs` yamls would just be `=3.9` (or if we wanted to encode intent, `>=3.9, <3.10`). We can re-evaluate this at each release since the CI yamls are usually where we look to determine packaging requirements, so we'll be motivated to resolve the `Mapping` thing every time we have to downpin to `python<3.10` in the package requirements.

I don't think we should have a separate yaml for local development builds - it seems really likely that it'd fall out of sync and we'd stop using it.

j-wags: After some thought, Matt's convinced me that it's best to only specify python version pins in the workflow file, and not in the conda-envs yaml. While it would solve the "building a dev environment gives me a broken env with python 3.10" problem _now_, it will create a problem in just a few hours when we're trying to fix python 3.10 compatibility and the pin in our conda-envs yaml is actively PREVENTING us from making a dev environment with python>=3.10. 

Neither solution is perfect, but incompatibility with the newer versions of python should be a rare occurrence, and it'll be USEFUL if dev builds warn us about incompatibilities with new python versions, so we can get them onto the issue tracker ASAP.

ntBre: In that case, should Python 3.10 also be specified in the workflow YAML? I see in the workflow Matt linked from the openff-forcefields repo that the Python version matrix contains both Python 3.9 and 3.10, but in this repo it only has 3.9. That would probably be a better way to catch incompatibilities than someone installing a new dev environment. And Python 3.11 was release 10 months ago, are there any plans for supporting/checking that too?

mattwthompson: The goal should be to support and test all supported versions within NEP 29, which currently includes 3.9-3.11. I have no idea if this will work on each of those versions right now

https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table

ntBre: Sounds like we should add 3.10 and 3.11 to the workflow then.

j-wags: > In that case, should Python 3.10 also be specified in the workflow YAML?

Yeah, with more dev time + CI resources we'd have all of our packages tested on 3.9, 3.10, and 3.11.

j-wags (approved):
This looks great - Thanks @ntBre @jthorton and @mattwthompson for pitching in/advising here. 

I can't figure out the codecov stuff - 

* There shouldn't need to be a secret token since this is a public repo
* I can visually confirm that some tests are running, so SOME source code lines should be getting covered, which isn't acknowledged on codecov
* Codecov acknowledges receipt of a coverage report, it just explicitly reports 0 coverage on every line
* Codecov has historically been flaky so maybe this is just getting confused by a merge conflict and getting the changes into the main branch will straighten it out. 

If the codecov stuff doesn't resolve, we can handle it in a separate PR. So this is good to merge pending green CI!

mattwthompson:
> There shouldn't need to be a secret token since this is a public repo

You're right, and this is what the documentation says. But what the documentation doesn't say is that without using an auth token, they sometimes get rate-limited by GitHub because they just pull down the stuff using the public interface.

Anyway, let me kick the tires on one of [my classic blind attempts](https://github.com/openforcefield/openff-nagl/pull/44/files) in case it magically fixes things

mattwthompson:
Well, I don't know if "sprinkle in some `python -m `" will fix my car, but it sure seems to do the trick here, ...

j-wags:
giphy-2657520286

PR #228, opened 2023-08-09
Make loading datasets skip over corrupt entries with invalid CMILES

### LINKED ISSUE #216: QCSubmit cannot process some existing datasets
```python
from openff.qcsubmit.results import OptimizationResultCollection
from qcportal import FractalClient

client = FractalClient()

OptimizationResultCollection.from_server(
    client=client,
    datasets="OpenFF Industry Benchmark Season 1 v1.0",
    spec_name="default",
)
```
Out
```
<<<Several parchment scrolls of stereochemistry warnings>>>

ValueError: 'hydrogens_are_explicit' was specified as True, but OpenEye Toolkit interpreted SMILES '[NH+:1]([C@:2]([C:3](=[O:4])[O-:10])([C:5]([C:6]([C:7]([C:8]([NH+:9]([H:22])[H:23])([H:20])[H:21])([H:18])[H:19])([H:16])[H:17])([H:14])[H:15])[H:13])([H:11])[H:12]' as having implicit hydrogen. If this SMILES is intended to express all explicit hydrogens in the molecule, then you should construct the desired molecule as an OEMol (where oechem.OEHasImplicitHydrogens(oemol) returns False), and then use Molecule.from_openeye() to create the desired OFFMol.
```
This appears to result from newer (0.11+?) versions of the toolkit having different cheminformatics assumptions. There doesn't appear to be a documented way to filter records _while_ grabbing a collection from the server:

```python
In [8]: OptimizationResultCollection.from_server?
Signature:
OptimizationResultCollection.from_server(
    client: qcportal.client.FractalClient,
    datasets: Union[str, Iterable[str]],
    spec_name: str = 'default',
) -> 'OptimizationResultCollection'
Docstring:
Retrieve (and deduplicate) the COMPLETE record ids referenced by the
specified datasets.

Args:
    client: The fractal client that should be used to interface with the running
        QCFractal instance which stores the datasets and their associated
        results records.
    datasets: The names of the datasets
    spec_name: The name of the QC specification that the records to retrieve
        should have been computed using.

Returns:
    A results collection object containing the record ids and a minimal amount
    of associated metadata such as the CMILES of the associated molecule.
File:      ~/mambaforge/envs/openff-interchange-env/lib/python3.10/site-packages/openff/qcsubmit/results/results.py
Type:      method
```

Creating a separate environment with an older version of the toolkit, grabbing the collection, and serializing out to disk is a workaround. I argue this is not sustainable nor consistent with the ethos that our infrastructure should facilitate scientists accessing existing data with minimal overhead.

```python
In [9]: from openff.toolkit import __version__
   ...:
   ...: print(__version__)
0.13.1

In [10]: from openff.qcsubmit import __version__
    ...: print(__version__)
0.4.1
```

j-wags:
## Todos

  - [x] Fix #216
  - [x] Add tests
  - [ ] ~(optional) Make both backends give the same number of molecules when loading the industry dataset.~ (these changes need to go into the OpenFF Toolkit)
  - [x] Emit a warning with a dataset entry is skipped for having an invalid CMILES. 
  - [x] ~Open issue in OpenFF Toolkit about making RDKitToolkitWrapper be strict about mapped SMILES with implicit Hs~ https://github.com/openforcefield/openff-toolkit/issues/1696
  - [x] ~Open issue in OpenFF Toolkit about preventing such SMILES from being created again.~ https://github.com/openforcefield/openff-toolkit/issues/1697
  - [x] ~Open issue in qca-dataset-submission about the corrupt entries (not planned for immediate action, but as a place for people to collect requests for the creation of a cleaned dataset)~ https://github.com/openforcefield/qca-dataset-submission/issues/327

## Status
- [x] Ready to go

### DIFF
--- a/openff/qcsubmit/results/results.py
+++ b/openff/qcsubmit/results/results.py
@@ -3,6 +3,7 @@ A module which contains convenience classes for referencing, retrieving and filt
 results from a QCFractal instance.
 """
 import abc
+import warnings
 from collections import defaultdict
 from typing import (
     TYPE_CHECKING,
@@ -467,6 +468,10 @@ class OptimizationResultCollection(_BaseResultCollection):
                             allow_undefined_stereo=True,
                         )
                     except ValueError:
+                        warnings.warn(
+                            f"Skipping entry {entry.name} with invalid CMILES {entry.attributes['canonical_isomeric_explicit_hydrogen_mapped_smiles']}",
+                            UserWarning,
+                        )
                         continue
                     inchi_key = mol.to_inchikey(fixed_hydrogens=True)
 
--- a/openff/qcsubmit/results/results.py
+++ b/openff/qcsubmit/results/results.py
@@ -441,26 +441,36 @@ class OptimizationResultCollection(_BaseResultCollection):
             client = dataset.client
             query = dataset.query(spec_name)
 
-            result_records[client.address].update(
-                {
-                    query[entry.name].id: OptimizationResult(
-                        record_id=query[entry.name].id,
-                        cmiles=entry.attributes[
-                            "canonical_isomeric_explicit_hydrogen_mapped_smiles"
-                        ],
-                        inchi_key=entry.attributes.get("fixed_hydrogen_inchi_key")
-                        or Molecule.from_mapped_smiles(
-                            entry.attributes[
-                                "canonical_isomeric_explicit_hydrogen_mapped_smiles"
-                            ],
-                            allow_undefined_stereo=True,
-                        ).to_inchikey(fixed_hydrogens=True),
-                    )
-                    for entry in dataset.data.records.values()
-                    if entry.name in query
-                    and query[entry.name].status.value.upper() == "COMPLETE"
-                }
-            )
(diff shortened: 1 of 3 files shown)


jthorton:
@j-wags Thanks for sorting this one, I have added a mocked dataset with an invalid entry which should test this case. Thinking about the other dataset types would we want to extend this fix to those as well?

jthorton:
> In its current form this keeps the OE backend from crashing on bad entries, but loading the dataset using OE vs. RDKit will yield a different number of molecules. 

Good point should this be an issue on the toolkit that it should raise an error when trying to create a molecule from an incomplete mapped smiles?

j-wags:
Yeah, after some thought the only pathway to consistently handling the corrupt molecules with both backends is to make changes in RDKitToolkitWrapper in the OFF Toolkit. 

I've added some to-dos above - Most of them are bookkeeping and followups, but one thing I'm going to do here is add a warning when a dataset entry is skipped for having an invalid CMILES.

j-wags:
Thanks for the huge assist @jthorton!

PR #229, opened 2023-08-10
Update docs

ntBre:
## Description
Update the docs conda environment to build the docs successfully, and add a couple of examples/links to the docs themselves.

## Todos
  - [x] update env
  - [x] update docs

## Status
- [x] Ready to go

### DIFF
--- a/devtools/conda-envs/docs.yaml
+++ b/devtools/conda-envs/docs.yaml
@@ -19,7 +19,7 @@ dependencies:
   - sphinxcontrib-bibtex
   - sphinx-notfound-page
   - myst-parser
-  - autodoc_pydantic
+  - autodoc-pydantic
   - pandoc
   - ipython
 
--- a/devtools/conda-envs/docs.yaml
+++ b/devtools/conda-envs/docs.yaml
@@ -25,7 +25,7 @@ dependencies:
 
     # Standard dependencies
   - openff-toolkit-base =0.14.0
-  - rdkit                #MOCKED
+  - rdkit
   - pydantic =1
   - pyyaml
   - qcportal >=0.15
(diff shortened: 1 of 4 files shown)


devtools/conda-envs/docs.yaml
```diff
@@ -19,13 +19,14 @@ dependencies:
   - sphinxcontrib-bibtex
   - sphinx-notfound-page
   - myst-parser
+  - autodoc_pydantic
```
lilyminium, line 22:
```suggestion
  - autodoc-pydantic
```

I think the spelling is different on conda! https://anaconda.org/conda-forge/autodoc-pydantic

ntBre: Thanks Lily

j-wags (approved):
Nice @ntBre - Thanks for catching this and getting everything going again! The `inherited-members` change seems good for this method, though I don't have time this morning to see whether it's forwarding anything that we don't intend elsewhere in the docs (I think the odds are low, and internal folks are the primary consumers of these docs anyway, so we can just roll it back if something weird is happening).

Thanks!

devtools/conda-envs/docs.yaml
```diff
@@ -19,13 +19,14 @@ dependencies:
...
+  - autodoc-pydantic
   - pandoc
   - ipython
 
     # Standard dependencies
-  - openff-toolkit-base >= 0.11.2
-#  - rdkit                MOCKED
-  - pydantic
+  - openff-toolkit-base =0.14.0
+  - rdkit                #MOCKED
```
j-wags, line 28:
```suggestion
  - rdkit
```

PR #231, opened 2023-08-15
Update QCSubmit for removal of ChemicalEnvironment class in OFFTK >0.14.1

ntBre:
## Description
[openff-toolkit 0.14.1](https://docs.openforcefield.org/projects/toolkit/en/stable/releasehistory.html#api-breaking-changes) removed the deprecated ChemicalEnvironment class and its associated module. We only used this in one place (plus a re-export of an exception type). This PR cleans up that one call site as suggested by @j-wags.

## Questions
- [ ] Do we want to keep `check_environments` at all?

## Status
- [ ] Ready to go

### DIFF
--- a/openff/qcsubmit/validators.py
+++ b/openff/qcsubmit/validators.py
@@ -296,7 +296,11 @@ def check_environments(environment: str) -> str:
         # capability "find_smarts_matches" for args...' it would be nice for
         # chemical_environment_matches to raise a more specific exception, but
         # it just raises a ValueError
-        if 'capability "find_smarts_matches"' not in str(e):
+        s = str(e)
+        if (
+            'capability "find_smarts_matches"' not in s
+            or "Available toolkits are: []" in s
+        ):
             raise e
 
     # we've either already returned successfully, raised an unrelated
--- a/openff/qcsubmit/validators.py
+++ b/openff/qcsubmit/validators.py
@@ -5,11 +5,9 @@ import re
 from typing import List, Tuple, Union
 
 import qcelemental as qcel
+from openff.toolkit import Molecule
 from openff.toolkit import topology as off
-from openff.toolkit.typing.chemistry.environment import (
-    ChemicalEnvironment,
-    SMIRKSParsingError,
-)
+from openff.toolkit.utils.exceptions import SMIRKSParsingError
 
 from openff.qcsubmit.constraints import Constraints
 from openff.qcsubmit.exceptions import (
@@ -288,14 +286,22 @@ def check_environments(environment: str) -> str:
     """
 
     # try and make a new chemical environment checking for parse errors
-    _ = ChemicalEnvironment(smirks=environment)
+    try:
+        _ = Molecule().chemical_environment_matches(environment)
+        # check for numeric tags in the environment
+        if re.search(":[0-9]]+", environment) is not None:
+            return environment
+    except ValueError as e:
+        # only catch an error like 'No registered toolkits can provide the
+        # capability "find_smarts_matches" for args...' it would be nice for
+        # chemical_environment_matches to raise a more specific exception, but
+        # it just raises a ValueError
+        if 'capability "find_smarts_matches"' not in str(e):
+            raise e
 
-    # check for numeric tags in the environment
-    if re.search(":[0-9]]+", environment) is not None:
-        return environment
-
-    else:
-        raise SMIRKSParsingError(
-            "The smarts pattern passed had no tagged atoms please tag the atoms in the "
-            "substructure you wish to include/exclude."
-        )
+    # we've either already returned successfully, raised an unrelated
+    # exception, or failed to parse the smirks
+    raise SMIRKSParsingError(
+        "The smarts pattern passed had no tagged atoms please tag the atoms in the "
+        "substructure you wish to include/exclude."
+    )
--- a/devtools/conda-envs/basic.yaml
+++ b/devtools/conda-envs/basic.yaml
@@ -27,7 +27,7 @@ dependencies:
 
   # openmmforcefields brings in the full toolkit; if that is ever re-packaged
   # this should be changed to openff-toolkit-base
-  - openff-toolkit == 0.14.0
+  - openff-toolkit >= 0.14.0
   - openff-units >=0.2.1
   - pydantic =1
   - pyyaml
--- a/devtools/conda-envs/psi4.yaml
+++ b/devtools/conda-envs/psi4.yaml
@@ -28,7 +28,7 @@ dependencies:
 
   ### Core dependencies.
 
-  - openff-toolkit-base == 0.14.0
+  - openff-toolkit-base >= 0.14.0
   - openff-units >=0.2.1
   - rdkit
   - pydantic =1
--- a/openff/qcsubmit/tests/test_workflow_components.py
+++ b/openff/qcsubmit/tests/test_workflow_components.py
@@ -901,6 +901,14 @@ def test_smarts_filter_validator():
         # good smarts with no tagged atoms.
         workflow_components.SmartsFilter(allowed_substructures=["[C]=[C]"])
 
+    from openff.toolkit.utils import ToolkitRegistry
+    from openff.toolkit.utils.toolkit_registry import _toolkit_registry_manager
+
+    # this test is the same as above, but without any toolkit available
+    with pytest.raises(ValueError):
+        with _toolkit_registry_manager(ToolkitRegistry()):
+            workflow_components.SmartsFilter(allowed_substructures=["[C]=[C]"])
+
     # a good search string
     smart_filter = workflow_components.SmartsFilter(
         allowed_substructures=["[C:1]=[C:2]"]
--- a/openff/qcsubmit/tests/test_workflow_components.py
+++ b/openff/qcsubmit/tests/test_workflow_components.py
@@ -877,7 +877,7 @@ def test_smarts_filter_validator():
     Make sure the validator is checking the allowed and filtered fields have valid smirks strings.
     """
 
-    from openff.toolkit.typing.chemistry import SMIRKSParsingError
+    from openff.toolkit.utils.exceptions import SMIRKSParsingError
 
     with pytest.raises(ValidationError):
         workflow_components.SmartsFilter(


openff/qcsubmit/validators.py
```diff
@@ -288,14 +286,22 @@ def check_environments(environment: str) -> str:
...
+    try:
+        _ = Molecule().chemical_environment_matches(environment)
+        # check for numeric tags in the environment
+        if re.search(":[0-9]]+", environment) is not None:
+            return environment
+    except ValueError as e:
+        # only catch an error like 'No registered toolkits can provide the
+        # capability "find_smarts_matches" for args...' it would be nice for
+        # chemical_environment_matches to raise a more specific exception, but
+        # it just raises a ValueError
```
mattwthompson, line 298: I'm not sure this is the same `ValueError` you're working into here, but updating the toolkit to raise more specific exceptions is a long-standing goal of ours - I opened https://github.com/openforcefield/openff-toolkit/pull/1695 after reading this comment but not before re-reading it again and being less sure it's the specific exception that is relevant to this bubble-up

ntBre: That looks like the same error to me! `Molecule.chemical_environment_matches` calls `toolkit_registry.call("find_smarts_matches", ...)` which calls `_find_smarts_matches` as far as I can tell, so hopefully that's the right place.

ntBre: It felt pretty hacky to inspect the exception as a string, but I didn't want to except anything, or even just any ValueError, which is why I added the comment too.

mattwthompson: Heh, it's not truly hacky until you're reaching into dunder methods and monkeypatching the stdlib.

What confuses me here is that `ValueError` can also be raised by the general machinery of how the `Molecule` class calls out to `ToolkitRegistry`s, particularly that it's raised when it tries stuff and everything fails:

```python
In [1]: from openff.toolkit import Molecule

In [2]: molecule = Molecule.from_smiles("ClC(Cl)(Cl)Cl", allow_undefined_stereo=True)

In [3]: molecule.assign_partial_charges(partial_charge_method="hartree-fock")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[3], line 1
----> 1 molecule.assign_partial_charges(partial_charge_method="hartree-fock")

File ~/mambaforge/envs/openff-interchange-env/lib/python3.11/site-packages/openff/toolkit/topology/molecule.py:2598, in FrozenMolecule.assign_partial_charges(self, partial_charge_method, strict_n_conformers, use_conformers, toolkit_registry, normalize_partial_charges)
   2584         warnings.warn(
   2585             f"Warning! Partial charge method '{partial_charge_method}' is not designed "
   2586             "for use on large (i.e. > 150 atoms) molecules and may crash or take hours to "
   (...)
   2590             stacklevel=2,
   2591         )
   2593 if isinstance(toolkit_registry, ToolkitRegistry):
   2594     # We may need to try several toolkitwrappers to find one
   2595     # that supports the desired partial charge method, so we
   2596     # tell the ToolkitRegistry to continue trying ToolkitWrappers
   2597     # if one raises an error (raise_exception_types=[])
-> 2598     toolkit_registry.call(
   2599         "assign_partial_charges",
   2600         molecule=self,
   2601         partial_charge_method=partial_charge_method,
   2602         use_conformers=use_conformers,
   2603         strict_n_conformers=strict_n_conformers,
   2604         normalize_partial_charges=normalize_partial_charges,
   2605         raise_exception_types=[],
   2606         _cls=self.__class__,
   2607     )
   2608 elif isinstance(toolkit_registry, ToolkitWrapper):
   2609     toolkit_wrapper: ToolkitWrapper = toolkit_registry

File ~/mambaforge/envs/openff-interchange-env/lib/python3.11/site-packages/openff/toolkit/utils/toolkit_registry.py:370, in ToolkitRegistry.call(self, method_name, raise_exception_types, *args, **kwargs)
    368 for toolkit, error in errors:
    369     msg += " {} {} : {}\n".format(toolkit, type(error), error)
--> 370 raise ValueError(msg)

ValueError: No registered toolkits can provide the capability "assign_partial_charges" for args "()" and kwargs "{'molecule': Molecule with name '' and SMILES 'C(Cl)(Cl)(Cl)Cl', 'partial_charge_method': 'hartree-fock', 'use_conformers': None, 'strict_n_conformers': False, 'normalize_partial_charges': True, '_cls': <class 'openff.toolkit.topology.molecule.Molecule'>}"
Available toolkits are: [ToolkitWrapper around OpenEye Toolkit version 2023.1.0, ToolkitWrapper around The RDKit version 2023.03.2, ToolkitWrapper around AmberTools version 22.0, ToolkitWrapper around Built-in Toolkit version None]
 ToolkitWrapper around OpenEye Toolkit version 2023.1.0 <class 'openff.toolkit.utils.exceptions.ChargeMethodUnavailableError'> : partial_charge_method 'hartree-fock' is not available from OpenEyeToolkitWrapper. Available charge methods are ['am1bcc', 'am1-mulliken', 'gasteiger', 'mmff94', 'am1bccnosymspt', 'am1elf10', 'am1bccelf10']
 ToolkitWrapper around The RDKit version 2023.03.2 <class 'openff.toolkit.utils.exceptions.ChargeMethodUnavailableError'> : partial_charge_method 'hartree-fock' is not available from RDKitToolkitWrapper. Available charge methods are ['mmff94', 'gasteiger']
 ToolkitWrapper around AmberTools version 22.0 <class 'openff.toolkit.utils.exceptions.ChargeMethodUnavailableError'> : partial_charge_method 'hartree-fock' is not available from AmberToolsToolkitWrapper. Available charge methods are ['am1bcc', 'am1-mulliken', 'gasteiger']
 ToolkitWrapper around Built-in Toolkit version None <class 'openff.toolkit.utils.exceptions.ChargeMethodUnavailableError'> : Partial charge method "hartree-fock"" is not supported by the Built-in toolkit. Available charge methods are ['zeros', 'formal_charge']
```

In this case I just lack context to grok which one is being captured here - comparing against the stringified exception seems fine here for deciphering what failed.

lilyminium: Same -- I would assume that a ValueError is *always* raised with the `'capability "find_smarts_matches"'` error message here, because if I understand correctly a ValueError inside a toolkit `find_smarts_matches` should also trigger this outer toolkit registry error, right?

openff/qcsubmit/validators.py
```diff
@@ -288,14 +286,22 @@ def check_environments(environment: str) -> str:
...
+        if 'capability "find_smarts_matches"' not in str(e):
+            raise e
```
lilyminium, line 300: Currently, this means that if we don't have any valid toolkits installed, the function falls down to the `SMIRKSParsingError` below about not having any tagged atoms. Would it make sense to still raise this toolkit capability error to communicate that scenario?

ntBre: Hmm, I see what you mean. Is there a better way to differentiate between them than checking this string? Ideally I would only except SMIRKSParsingError, but if the underlying code doesn't differentiate I wasn't sure how to do better than this. We could certainly log a warning at the very least.

openff/qcsubmit/validators.py
mattwthompson: That sounds right to me - 

* If no toolkit can provide it, the API raises the "outer" `ValueError` with something that matches that string
* If toolkits can provide it but fail to parse the SMARTS, the "inner" error will be raised but the "outer" error will ultimately be the one that's raised

What I got hung up on is if this code (here) is trying to catch the "inner" error or not; if it needs to catch the inner error, my PR to the toolkit would be needed. But I think the "outer" error has the same result here, not sure

ntBre: I didn't understand this earlier, but I see now. I was trying to catch the "inner" error actually due to the SMARTS parsing, but if it makes sense to catch all of them (I'm not sure it does, no toolkit at all seems like a time to raise an exception here), I'd be open to that too. It sounds to me like the toolkit PR is needed to differentiate between the cases.

ntBre: Is it possible to write a test that runs without a toolkit registry (temporarily empties it maybe?) to check both cases?

mattwthompson: I'd try to use [this](https://github.com/openforcefield/openff-toolkit/blob/615320fe3497559c709fdb80c2cbf350511b21b3/openff/toolkit/utils/toolkit_registry.py#L379) context manager with an empty `ToolkitRegistry`, similar to how it's used [here](https://github.com/openforcefield/openff-toolkit/blob/0.14.3/openff/toolkit/typing/engines/smirnoff/forcefield.py#L1222-L1230) to wire through a user-defined registry to Interchange

ntBre: I'll give that a try tomorrow, thank you!

ntBre:
Alright, I added a test showing the ValueError (mis)behavior we discussed and then made another check against the exception string to ignore the no toolkit case. I don't really think adding more conditions to the except is a good idea, but is there anything else to do? Does this change just need to wait on Matt's PR and for it to make its way into a toolkit release?

openff/qcsubmit/validators.py
j-wags: This conversation is hitting on one of the weirdest (but as far as I can tell necessary) design points of the OFF Toolkit - The `ToolkitRegistry.call(..., raise_exception_types=)` kwarg. 

The `raise_exception_types` kwarg has basically three outcomes:
* `raise_exception_types=[Exception]` tries the first ToolkitWrapper that provides the correctly-named method, and either returns successfully or raises its exception verbatim
* `raise_exception_types=[]` tries, in order, each ToolkitWrapper that provides the correctly-named method. If one of them succeeds, it returns immediately. If one raises an exception, it holds onto that exception in a list. If all of them fail, a `ValueError` is raised with all of the exceptions from the list.
* `raise_exception_types=[SMIRKSParsingError]` tries, in order, each ToolkitWrapper that provides the correctly-named method. If one of them returns successfully then it immediately returns. If one of them raises an SMIRKSParsingError, it immediately raises that verbatim. If one of them raises another sort of exception, it holds on to it in case nothing works, and then includes it in the final raised `ValueError`

In the case of `Molecule.chemical_environment_matches`, [it uses `return_exception_types=[]`](https://github.com/openforcefield/openff-toolkit/blob/615320fe3497559c709fdb80c2cbf350511b21b3/openff/toolkit/topology/molecule.py#L3543). This is the second option above, so it will either return successfully or return a ValueError containing a list of other exceptions. So Matt's changes in https://github.com/openforcefield/openff-toolkit/pull/1695 are still an important cleanup but won't affect this PR (since even a more specific exception type will still be rolled into a ValueError). This is nice in the short run because it means these PRs aren't necessarily interdependent (which is good because it's one less potential compatibility issue to track).

openff/qcsubmit/validators.py
```diff
@@ -288,14 +286,26 @@ def check_environments(environment: str) -> str:
...
-    else:
-        raise SMIRKSParsingError(
-            "The smarts pattern passed had no tagged atoms please tag the atoms in the "
-            "substructure you wish to include/exclude."
-        )
+    try:
+        _ = Molecule().chemical_environment_matches(environment)
+        # check for numeric tags in the environment
+        if re.search(":[0-9]]+", environment) is not None:
+            return environment
```
j-wags, line 293: (not blocking, maybe future issue) I don't understand why we have this check. From what I understand the purpose of the SMIRKS filter in QCSubmit to be - that is, returning True or False on entire molecules - A SMIRKS without numeric tags should be valid.

j-wags (approved):
This looks great to me in its current form. As far as I can tell this doesn't interact with https://github.com/openforcefield/openff-toolkit/pull/1695 and I'm happy with the current state. Thanks @ntBre!

openff/qcsubmit/validators.py
ntBre: Yeah I'm not sure about that one either. I just carried it over because it was there before! Did you decide we do want to keep this check for now too? I left that in the Questions section at the top too.

j-wags: I decided to keep it. It's making this check slightly more strict then it needs to be, but

* This entire check is just a performance hack/UX improvement - without this check an invalid SMIRKS would just error out later when it's actually being used
* There may be some cases that I'm unaware of where removing this would break something, and it's possible that those cases aren't tested so CI wouldn't save us.
* The possibly-excessive requirement must not be too burdensome on users because nobody is complaining about it

j-wags: > From what I understand the purpose of the SMIRKS filter in QCSubmit to be - that is, returning True or False on entire molecules

Ah, I see, turns out this assumption is wrong. This method is used in [several places](https://github.com/openforcefield/openff-qcsubmit/blob/2461c47b63be2a5c328caa688bb59aa07b402be3/openff/qcsubmit/workflow_components/fragmentation.py#L25-L39) where there must be tagged atoms. So removing this would have been a regression. Good thing we didn't do that :-)

PR #238, opened 2023-11-08
Make OpenMM an optional dependency

mattwthompson:
## Description
The `simtk` namespace is now more than two years deprecated; it's long past a point in which it makes sense to worry about backwards compatibility for future releases.

Grepping through just instances of `simtk` also surfaced some cases in which things were probably using the wrong registry. I'll see how the tests behaves.

## Todos
  - [x] Remove `simtk` import guards

## Questions
- [ ] Question1

## Status
- [ ] Ready to go


openff/qcsubmit/datasets/datasets.py
```diff
@@ -497,10 +497,7 @@ def _get_missing_basis_coverage(
 
         import basis_set_exchange as bse
 
-        try:
-            from openmm.app import Element
-        except ImportError:
-            from simtk.openmm.app import Element
+        from openff.qcsubmit.validators import SYMBOLS_TO_ELEMENTS
```
dotsdl, line 500: :raised_hands:

mattwthompson:
Bit of a mistake at my part - I spent a chunk of time wondering why I couldn't reproduce the earlier CI failures locally ... only to notice I just hadn't pushed my recent changes. 🤦‍♂️

mattwthompson:
This is ready to go % #245

j-wags (approved):
Awesome, thanks @mattwthompson! I'll update the releasenotes and merge shortly.

PR #240, opened 2023-11-10
Expose ability to set OptimizationProtocols for for optimizations, torsiondrives

dotsdl:
## Description
[`OptimizationProtocols`](https://github.com/MolSSI/QCElemental/blob/1ef6b52adc4a18cfd6a2cfc1b56880daaeea3a44/qcelemental/models/procedures.py#L32-L48) allow users to e.g. only save gradients, wavefunctions for starting and final geometries of an optimization, reducing storage requirements for results. The changes here expose this functionality through QCSubmit, allowing users to set `OptimizationProtocols` settings other than the defaults if desired.

## Todos
  - [x] add tests

## Status
- [x] Ready to go

### DIFF
--- a/openff/qcsubmit/datasets/datasets.py
+++ b/openff/qcsubmit/datasets/datasets.py
@@ -20,7 +20,7 @@ import qcportal as ptl
 from openff.toolkit import topology as off
 from pydantic import Field, constr, validator
 from qcelemental.models import AtomicInput, OptimizationInput
-from qcelemental.models.procedures import QCInputSpecification
+from qcelemental.models.procedures import QCInputSpecification, OptimizationProtocols
 from qcportal import PortalClient, PortalRequestError
 from qcportal.optimization import OptimizationDatasetNewEntry, OptimizationSpecification
 from qcportal.singlepoint import (
@@ -907,7 +907,11 @@ class OptimizationDataset(BasicDataset):
         GeometricProcedure(),
         description="The optimization program and settings that should be used.",
     )
+    protocols: OptimizationProtocols = Field(
+            OptimizationProtocols(),
+            "Protocols regarding the manipulation of a Optimization output data.")
     dataset: Dict[str, OptimizationEntry] = {}
+
 
     @classmethod
     def _entry_class(cls) -> Type[OptimizationEntry]:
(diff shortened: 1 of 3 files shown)


openff/qcsubmit/datasets/datasets.py
```diff
@@ -907,6 +907,10 @@ class OptimizationDataset(BasicDataset):
         GeometricProcedure(),
         description="The optimization program and settings that should be used.",
     )
+    protocols: OptimizationProtocols = Field(
+        OptimizationProtocols(),
+        description="Protocols regarding the manipulation of a Optimization output data.",
```
lilyminium, line 912:
```suggestion
        description="Protocols regarding the manipulation of Optimization output data.",
```

dotsdl: Thanks! Got it!

openff/qcsubmit/tests/test_datasets.py
```diff
@@ -2610,6 +2611,25 @@ def test_optimizationdataset_qc_spec():
...
+        description="XXXXXXXX",
+    )
+
+    assert dataset.protocols.trajectory == "all"
+
+    # change the protocol to only save gradient evaluations for final state
+    dataset.protocols = OptimizationProtocols(trajectory="final")
+
+    assert dataset.protocols.trajectory == "final"
+
```
lilyminium, line 2631: Would it be worth also adding a test to `test_submissions` that actually carries out the calculation and checks that only the final gradients are saved, or would we expect that to be covered by the QC* stack so long as our inputs have the keyword?

dotsdl: I think it makes sense to do this. I'll modify the existing `test_optimization_submissions` to use `'final'` for the `trajectory` protocol and see if this gives the expected result.

dotsdl: Added!

lilyminium (approved):
Thanks for adding the protocol and test @dotsdl -- everything LGTM pending green CI!

dotsdl:
Thanks @lilyminium!

PR #246, opened 2023-12-01
Skip slow test on macOS

mattwthompson:
Band-aid for https://github.com/openforcefield/openff-qcsubmit/issues/245

This is not a fix, but a way to get tests passing and unblock other PRs. The test seems to be passing on Linux, so I'm not overly worried.

## Description
Skip slow test on macOS

## Status
- [x] Ready to go


j-wags:
A better fix would be to switch this to use a smaller molecule. If we skip this on MacOS we'll forgot and never come back to it. In the short term, I'd prefer to just bump the max_iter in the await_services call even higher.

mattwthompson:
300 iterations is already pretty high, no?

j-wags:
Feel free to switch to using a very small molecule with just enough torsions to reproduce the test (like... some short alcohol or something from SMILES). And bonus points if you make it as an SDF instead of a mol2.

j-wags:
Superseded by #252

PR #251, opened 2023-12-15
Use Psi4 1.9

mattwthompson:
## Description
Upgrade to Psi4 1.9, which might not need the custom Libint label.

## Todos
  - [x] Upgrade to Psi4 1.9
  - [x] Drop Libint channel

## Status
- [x] Ready to go


j-wags (approved):
LGTM. I fretted over whether this would need corresponding changes in a bunch of places in the install docs but... the [current install docs](https://docs.openforcefield.org/projects/qcsubmit/en/stable/installation.html) don't mention psi4! Simplicity begets simplicity, it seems. Thanks @mattwthompson!

Issue #255, opened 2024-01-30
Deduplication inchi check vs aromaticity


lilyminium:
The deduplication check hashes molecules via inchikey, then tries to remap them via an isomorphism check. Unfortunately structures like the below have the same inchikey, but different bond orders/aromaticity perception, so this raises an assertion error here: https://github.com/openforcefield/openff-qcsubmit/blob/2fa465a87ffea4c5b02d5d0cd8622d1a22d4fe83/openff/qcsubmit/workflow_components/utils.py#L713-L723

In this particular case, adding `aromatic_matching=False` to the `are_isomorphic` check did fix it for me.

<img width="269" alt="Screenshot 2024-01-30 at 8 51 13 pm" src="https://github.com/openforcefield/openff-qcsubmit/assets/31115101/1ce0a966-994b-48eb-ab27-ee27ce301792">
<img width="283" alt="Screenshot 2024-01-30 at 8 51 16 pm" src="https://github.com/openforcefield/openff-qcsubmit/assets/31115101/07c89a0a-7054-43e0-8a37-0d5157f0a27d">

jthorton:
Ah another one for the ever-growing list of deduplication edge cases which is where the formal charge and bond order matching settings come from. I think your solution should be fine generally as this check is only ran if the inchi matches!

PR #260, opened 2024-01-30
Add `aromatic_matching=False` to isometry check

mattwthompson:
## Description

Taking @lilyminium's suggestion in #255 and just seeing what happens in tests.

Not squatting on this at all; if somebody wants to move this forward please feel free to use this branch/PR.

## Todos
  - [x] Add `aromatic_matching=False` to isometry check
  - [x] Add test reproducing #255

## Status
- [x] Ready to go

### DIFF
--- a/openff/qcsubmit/_tests/test_datasets.py
+++ b/openff/qcsubmit/_tests/test_datasets.py
@@ -343,6 +343,38 @@ def test_componentresult_deduplication_iso():
     assert result.n_conformers == 2
 
 
+def test_componentresult_deduplication_tautomer():
+    """Reproduce issue #260 using molecules in the reported issue."""
+
+    result = ComponentResult(
+        component_name="Test iso deduplication",
+        component_description={},
+        component_provenance={},
+    )
+
+    # these two molecules have the same InChI key but differ in aromaticity ...
+    mol1 = Molecule.from_smiles(
+        smiles="[H]c1c(c(c(c2c1N(C(=C2C3=C(N(C(=O)C(=C3[H])[H])C([H])([H])C([H])([H])[H])[H])[H])[H])[H])F)[H]",
+        hydrogens_are_explicit=True,
+    )
+    mol1.generate_conformers(n_conformers=1)
+
+    mol2 = Molecule.from_smiles(
+        smiles="[H]c1c(c([n+](c(c1C2=C(N(c3c2c(c(c(c3[H])[H])F)[H])[H])[H])[H])C([H])([H])C([H])([H])[H])[O-])[H]",
+        hydrogens_are_explicit=True,
+    )
+    mol2.generate_conformers(n_conformers=1)
+
+    # this function returns a boolean equal to "is this molecule already in this dataset"
+    assert not result.add_molecule(mol1)
+    assert not result.add_molecule(mol2)
+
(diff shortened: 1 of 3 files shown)


mattwthompson:
Just a note to myself: adding a test like `test_componentresult_deduplication_iso` (or some others in the same file) with an appropriate molecule should cover this change. IIUC the key behavior is that two molecules (or two representations of, or two tautomers of, depending on your view)

* have the same InChi key
* are viewed by the toolkit as isomorphic when `aromatic_matching=False`
* are not viewed by the toolkit as isomorphic when `aromatic_matching=True` (the current behavior, and the toolkit's default)

The molecule in the linked issue appears to trigger these conditions via having a proton hop around within an aromatic ring

openff/qcsubmit/_tests/test_datasets.py
```diff
@@ -343,6 +343,38 @@ def test_componentresult_deduplication_iso():
...
+    mol2.generate_conformers(n_conformers=1)
+
+    # this function returns a boolean equal to "is this molecule already in this dataset"
+    assert not result.add_molecule(mol1)
+    assert not result.add_molecule(mol2)
+
+    # ... and so the expected behavior is that they're deduplicated into the same
+    # "molecule" but the separate `Molecule` objects exist as distinct conformers
+    assert result.n_molecules == 1
+    assert result.n_conformers == 2
```
mattwthompson, line 375: I do not understand how this class and method are meant to behave, so everything I've written here should be fact-checked. This test is mostly cribbed from `test_componentresult_deduplication_iso ` above and the shifting boundaries of molecule identity make it hard to understand (at least for me at this particular moment). Brownie points if `ComponentResult.add_molecule` can be updated in a way that I understand what's going on

mattwthompson:
Test added! If I revert the `aromatic_matching=False` change this test fails; with the change, the test passes.

openff/qcsubmit/_tests/test_datasets.py
```diff
@@ -343,6 +343,38 @@ def test_componentresult_deduplication_iso():
     assert result.n_conformers == 2
 
 
+def test_componentresult_deduplication_tautomer():
+    """Reproduce issue #260 using molecules in the reported issue."""
```
j-wags, line 347:
```suggestion
    """Reproduce issue #255 using molecules in the reported issue."""
```

openff/qcsubmit/_tests/test_datasets.py
j-wags: I agree with you assessment that the second call to `add_molecule` should return `True`. Working on updating that now.

j-wags (approved):
Thanks for making the great test. I also had your confusion about the return value of `ComponentResult.add_molecule`, and I realized it was clearly a bug, since instead of returning a `bool`, it would return `None`, and `assert(None)` quacked like `assert(False)`. I've fixed that and now everything seems hunky dorey, and I've added a little more testing for it as well. I've updated the release notes saying as much.

PS in a previous commend you said 

> The molecule in the linked issue appears to trigger these conditions via having a proton hop around within an aromatic ring

and I'll clarify for future readers that it's a double bond/formal charge that's hopping around, not a proton!

jthorton:
Great job guys!

Issue #264, opened 2024-03-11
Psi4 tests failing


mattwthompson:
https://github.com/openforcefield/openff-qcsubmit/actions/runs/8225781166/job/22491290735

https://github.com/openforcefield/openff-qcsubmit/actions/runs/8234860096/job/22517602570

mattwthompson:
Looks like production is using 1.8.2: https://github.com/openforcefield/qca-dataset-submission/blob/ac79bc1fcc84b16db5e74fb3a0d3b9639c1b389e/devtools/prod-envs/qcarchive-worker-openff-psi4.yaml#L12

mattwthompson:
https://github.com/psi4/psi4/releases/tag/v1.9.1

j-wags:
Unable to reproduce natively on mac. The following are docker instructions to spin up a linux container and reproduce the error:

`docker run -it mambaorg/micromamba`

in another terminal

`docker cp ~/oe_license.txt <container_id>:/home/mambauser`

in container

```
micromamba install -c conda-forge git
git clone https://github.com/openforcefield/openff-qcsubmit.git
micromamba env create -f openff-qcsubmit/devtools/conda-envs/psi4.yaml -n qcsubmit-dev
cd openff-qcsubmit/
pip install -e .
OE_LICENSE=/home/mambauser/oe_license.txt pytest "openff/qcsubmit/_tests/test_submissions.py::test_basic_submissions_single_spec[PSI4 hf 3-21g energy]" 
```

Full output:

```
=================================== FAILURES ===================================
___________ test_basic_submissions_single_spec[PSI4 hf 3-21g energy] ___________

fulltest_client = PortalClient(server_name='QCFractal Server', address='http://127.0.0.1:39423/', username='None')
specification = ({'basis': '3-21g', 'method': 'hf', 'program': 'psi4'}, 'energy')

    @pytest.mark.parametrize(
        "specification",
        [
            pytest.param(
                ({"method": "hf", "basis": "3-21g", "program": "psi4"}, "energy"),
                id="PSI4 hf 3-21g energy",
            ),
            pytest.param(
                (
                    {
                        "method": "openff-2.1.0",
                        "basis": "smirnoff",
                        "program": "openmm",
                    },
                    "energy",
                ),
                id="SMIRNOFF openff-2.1.0 energy",
            ),
            pytest.param(
                ({"method": "uff", "basis": None, "program": "rdkit"}, "gradient"),
[...]

PR #265, opened 2024-03-11
Use Psi4 1.9.0

mattwthompson:
#264


mattwthompson:
Well that did not do the trick

PR #266, opened 2024-03-11
Update `pytest` arguments

mattwthompson:
Currently,

* `pytest-xdist` is installed but not leveraged
* test output is verbose for all tests, making it excessively long and difficult to parse

https://docs.pytest.org/en/8.0.x/how-to/output.html


mattwthompson:
#263 would be nice to get in before this

j-wags:
Unblocked by #263, but currently CI is failing for reasons that are likely on MolSSI's side. Given that xdist might cause its own concurrency issues, let's wait until we get green CI on main to test+merge this.

j-wags (approved):
LGTM!

Issue #267, opened 2024-03-12
`__all__` contains wrong types


mattwthompson:
```python
In [1]: from openff.qcsubmit.results import *
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[1], line 1
----> 1 from openff.qcsubmit.results import *

File <frozen importlib._bootstrap>:1073, in _handle_fromlist(module, fromlist, import_, recursive)

File <frozen importlib._bootstrap>:1069, in _handle_fromlist(module, fromlist, import_, recursive)

TypeError: Item in openff.qcsubmit.results.__all__ must be str, not ModelMetaclass
```

The problem is where `__all__` is defined as a [list of classes](https://github.com/openforcefield/openff-qcsubmit/blob/0.50.2/openff/qcsubmit/results/__init__.py#L10-L12) instead of strings, which is required

> The public names defined by a module are determined by checking the module’s namespace for a variable named __all__; if defined, it must be a sequence of strings which are names defined or imported by that module. 

https://docs.python.org/3/reference/simple_stmts.html#import

There are [a few other](https://github.com/search?q=repo%3Aopenforcefield%2Fopenff-qcsubmit%20__all__&type=code) such instances

PR #268, opened 2024-03-12
Ensure `__all__` contains only str

mattwthompson:
#267


j-wags (approved):
This looks great @mattwthompson, good catch. Test failures are unrelated. I've updated the releasenotes and will merge.

PR #270, opened 2024-03-20
Speed up TorsionDriveResultCollection.to_records by batching requests

ntBre:
## Description
These changes make the following code run in ~16 seconds instead of ~65 seconds on my computer by sending lists of record IDs to QCPortal instead of single records at a time. This should be safe to do because all three of `PortalClient.get_torsiondrives`, `PortalClient.get_optimizations`, and `PortalClient.get_molecules` state that they return their results in the same order as the given IDs. [[1](https://github.com/MolSSI/QCFractal/blob/ec8ca6a2eac73325724466e927858573e0b6efbc/qcportal/qcportal/client.py#L1422)][[2](https://github.com/MolSSI/QCFractal/blob/ec8ca6a2eac73325724466e927858573e0b6efbc/qcportal/qcportal/client.py#L1209)][[3](https://github.com/MolSSI/QCFractal/blob/ec8ca6a2eac73325724466e927858573e0b6efbc/qcportal/qcportal/client.py#L343)]

I also used the `include` kwarg on `get_torsiondrives`, which is supposed to pre-fetch those parts of the record instead of fetching them later when the properties are accessed, but because [_handle_includes](https://github.com/MolSSI/QCFractal/blob/ec8ca6a2eac73325724466e927858573e0b6efbc/qcportal/qcportal/torsiondrive/record_models.py#L228) appears to handle these on a per-record basis, this probably doesn't help the performance.

Profiling code:

```python
import cProfile
from pstats import SortKey

from openff.qcsubmit.results import TorsionDriveResultCollection

pr = cProfile.Profile()

with cProfile.Profile() as pr:
    client = PortalClient("https://api.qcarchive.molssi.org:443/")
    ds = "OpenFF Torsion Coverage Supplement v1.0"
    td = TorsionDriveResultCollection.from_server(client, ds)
    records = td.to_records()
    pr.print_stats(SortKey.CUMULATIVE)

assert len(records) == 40
```

## Todos
  - [x] TorsionDriveResultCollection
  - [x] OptimizationResultCollection

## Status
- [ ] Ready to go


lilyminium:
Just to comment on the speed, my laptop is clearly not as nice as your computer but I'm also seeing ~4x speedup: 260 seconds in `main` and 53 seconds in this branch. 🎉

bennybp:
> because [_handle_includes](https://github.com/MolSSI/QCFractal/blob/ec8ca6a2eac73325724466e927858573e0b6efbc/qcportal/qcportal/torsiondrive/record_models.py#L228) appears to handle these on a per-record basis, this probably doesn't help the performance.

The fix for that is coming in a future PR! So go ahead and use `include` and in the future it should be even faster

j-wags:
This is promising - I added a ticket for this on [our Trello](https://trello.com/b/dzvFZnv4/infrastructure) before I saw you hadn't checked some of the boxes in the description. Apologies for the false start @ntBre, let me know when this is ready for review!

ntBre:
@j-wags I should have just looked at this yesterday! The `OptimizationResultCollection` already had very similar code to what I added for the torsion drives. I just added `final_molecule` to the `include` part for the future. This should be ready for review!

j-wags:
Excellent. Looking at this with fresh eyes, since this touches some business logic there's a chance that the already-failing psi4 tests could mask a problem here. I'm going to look into those real fast but if I can't resolve them quickly I'll give this PR the benefit of the doubt and review today.

ntBre:
Sounds good! I should also mention that I added this code to the bottom of the profiling script above:

```python
with open("got.dat", "w") as out:
    for r, m in records:
        print(r.id, m.to_smiles(), file=out)
```

and confirmed that it was the same before and after the changes, in addition to the `assert` that was already there. I can try to add this as a real test, but I wasn't sure if we wanted to be making real requests in the tests. Matt said on slack that this functionality should be covered (somehow) by existing tests anyway, but it could overlap with the Psi4 tests, as you said.

j-wags (approved):
Alright, I've got CI passing and, as expected, this looks excellent. Thanks @ntBre! Please update the [releasenotes](https://github.com/openforcefield/openff-qcsubmit/blob/main/docs/releasehistory.md) before you merge. 

I've added a trello ticket to get this into a release, which I'll create once this is merged.

ntBre:
@j-wags Release notes updated, feel free to merge this if they look good to you!

PR #272, opened 2024-03-21
Fix install instructions

### LINKED ISSUE #271: meta.yaml missing
devtools/conda-env/meta.yaml mentioned in the installation  is missing from distribution. Can this be updated.

j-wags:
## Description
Our install docs were referring people to `devtools/conda-envs/meta.yaml`, which doesn't exist any more. We could either point them to `psi4.yaml` or `basic.yaml`. I think basic is better since it's more likely to be devs installing from source, and they'll know how to install psi4 afterwards if they need.

## Todos
  - [x] Fixes #271

## Status
- [ ] Ready to go

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -30,7 +30,7 @@ cd openff-qcsubmit
 Create a custom conda environment which contains the required dependencies and activate it:
 
 ```bash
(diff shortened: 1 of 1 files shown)


j-wags:
The test failures are expected and are due to a different issue.

Issue #275, opened 2024-04-09
QCSubmit CI is failing with QCF 0.54


j-wags:
Nightly CI has been failing for the last 3 days. Seems to be something with pydantic object parsing.

bennybp:
This looks to be something with the new QCPortal. It's not fetching all the fields required for some reason. I don't see much wrong with qcsubmit. Let me dig into it a bit

bennybp:
Ok I do see a few issues with the v0.54 release. The good news is that it is largely an artifact with how the new caching mechanism can result in caches shared between tests. So overall QCSubmit should largely work regardless of the failing tests.

Let me think of a way to more properly disable cache sharing during tests

(The `entry_points` errors from a few days ago are related to Python 3.9 compatibility with networkx which has already been fixed)

j-wags:
Ahh, that makes sense. Would it work if we had the clients we create in our tests have a temporary directory for their `cache_dir`, or set the `cache_max_size` to 0? 

(and not that it's relevant since the networkx issue is resolved, but since we follow [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table) we're not going to be testing 3.9 any more)

bennybp:
I've got a PR up that should fix these issues. Would it be possible to test this somehow before I make a release?

If it works, I can do a quick v0.54.1 release

j-wags:
At least one of the tests (`openff/qcsubmit/_tests/results/test_filters.py::test_torsion_drive_record_filter_apply[result_filter0-expected_ids0]`) fails even if it's run alone with QCP 0.54 (and it passes with 0.53), so I'm going to apply a metadata patch to the existing conda-forge packages.

j-wags:
Hm, on second thought the test failures all seem to be due to test fixtures having problems, so I'll hold off on the patch.

j-wags:
Resolved by #277

PR #276, opened 2024-04-10
Pin QCPortal to 0.53

mattwthompson:
#275 is causing false negatives, so I'd like to pin this the upstream versions that are known to be compatible


j-wags:
Oh wow, I didn't expect that to be so straightforward (I thought the bug was all serverside)

PR #277, opened 2024-04-10
Updates for QCFractal 0.54 and OFFTK 0.16

mattwthompson:
* [x] Fixes https://github.com/openforcefield/openff-qcsubmit/issues/275
* [x] Fixes https://github.com/openforcefield/openff-qcsubmit/issues/278


mattwthompson:
Okay ... I didn't pay attention to which tests are failing, but this seems to accomplish the goal of testing against that PR!

j-wags:
This fixes some of the test failures, but 3 remain:

```
=========================== short test summary info ============================
FAILED openff/qcsubmit/_tests/results/test_filters.py::test_torsion_drive_record_filter_apply[result_filter0-expected_ids0] - pydantic.v1.error_wrappers.ValidationError: 1 validation error for TorsiondriveRecord
initial_molecules_ids_
  extra fields not permitted (type=value_error.extra)
FAILED openff/qcsubmit/_tests/results/test_results.py::test_torsion_smirnoff_coverage - pydantic.v1.error_wrappers.ValidationError: 1 validation error for TorsiondriveRecord
initial_molecules_
  extra fields not permitted (type=value_error.extra)
FAILED openff/qcsubmit/_tests/test_submissions.py::test_invalid_cmiles[TorsiondriveDatasetFactory-TorsionDriveResultCollection] - Failed: DID NOT WARN. No warnings of type (<class 'UserWarning'>,) were emitted.
The list of emitted warnings is: [].
```

bennybp:
Two small correction in qcsubmit (which I can push if you would like):

```diff --git a/openff/qcsubmit/tests/results/__init__.py b/openff/qcsubmit/tests/results/__init__.py
index 8b97bc2..ff10614 100644
--- a/openff/qcsubmit/tests/results/__init__.py
+++ b/openff/qcsubmit/tests/results/__init__.py
@@ -180,7 +180,7 @@ def mock_torsion_drive_result_collection(
                             ),
                         ),
                     ),
-                    initial_molecules_ids_=[
+                    initial_molecules_ids=[
                         i + 1
                         for i in range(
                             molecules[address][int(entry.record_id) - 1].n_conformers
diff --git a/openff/qcsubmit/tests/results/test_results.py b/openff/qcsubmit/tests/results/test_results.py
index 9af59ad..fe062b7 100644
--- a/openff/qcsubmit/tests/results/test_results.py
+++ b/openff/qcsubmit/tests/results/test_results.py
@@ -406,7 +406,7 @@ def test_torsion_smirnoff_coverage(public_client, monkeypatch):
                         ),
                     ),
                     id=entry.record_id,
-                    initial_molecules_=[],
+                    initial_molecules=[],
                     status=RecordStatusEnum.complete,
                     is_service=False,
                     created_on=datetime.datetime(2022, 4, 21, 0, 0, 0),
```

j-wags:
Oh cool. I'll push those, one sec.

bennybp:
I added one more fix to qcportal. You will probably get one test failure here, so it will need to be re-run

bennybp:
The last error is a bit puzzling to me:

```
FAILED openff/qcsubmit/_tests/test_submissions.py::test_invalid_cmiles[TorsiondriveDatasetFactory-TorsionDriveResultCollection] - Failed: DID NOT WARN. No warnings of type (<class 'UserWarning'>,) were emitted.
The list of emitted warnings is: [].
```

It happens locally for me on both v0.53 and v0.54

j-wags:
I think this one is because we're failing to set an attribute on an entry (running the getter later doesn't return the changed value). Using the setter this way isn't a part of our core functionality, we just use it to set an invalid value to ensure an error is later raised. 

From [this test](https://github.com/openforcefield/openff-qcsubmit/blob/main/openff/qcsubmit/_tests/test_submissions.py#L1370-L1431)
```
    entries[0].attributes[
        "canonical_isomeric_explicit_hydrogen_mapped_smiles"
    ] = "[H:4][C:2](=[O:1])[OH:3]"
```

bennybp:
You probably need to update the cache with the modified entry (awkwardly since this is not particularly supported directly yet):

```
1416     entries[0].attributes[
1417         "canonical_isomeric_explicit_hydrogen_mapped_smiles"
1418     ] = "[H:4][C:2](=[O:1])[OH:3]"
1419    
1420     ds._cache_data.update_entries(entries)
```

But that still doesn't get me the warning. I think I see the cmiles propagate correctly, but still nothing.

j-wags:
Welp, I spent hours failing to reproduce anything locally, and finally gave up and ran your proposed change in CI. This looks good, and I'll switch this PR to install QC* 0.54.1 once you've made the release.

j-wags:
@mattwthompson Any ideas as to why flake8 is failing? We didn't change the file in question (`openff/qcsubmit/common_structures.py`) and there hasn't been a new flake8 package in [3 months](https://anaconda.org/conda-forge/flake8).

mattwthompson:
This one is obscure; the actual versions of everything used by `pre-commit` (either locally or the CI service) are hard-coded in the config file. There just happens to be something different if it's run with 3.12 vs. older versions. The actual check is from `pycodestyle` (`flake8` is really just a wrapper). I can't find a stub issue for this particular behavior, but I can get the failure in a 3.12 environment and I don't get it in a 3.10 environment

The templating triggering the failure looks kinda funky; there might be a different way to write it, but I don't want to poke the bear

```python
        pcm_string = f"""
        Units = {self.units}
        CODATA = {self.codata}
        Medium {{{medium_str
        }}}
        Cavity {{{cavity_str}}}"""

j-wags:
Yeah, agreed. The code in question looks properly weird to me so I don't want to touch it haha. Also it's weird that flake8 is complaining about something _inside_ a string, but I guess it's a f-string so the jurisdiction is muddled there.

j-wags:
Thanks for the input. This is the kind of low-stakes rabbit-hole decision that usually consumes all my mental power, so instead of doing that I've flipped a coin and it decided that I'll just merge this with Matt's changes to the pre-commit config and without changing the f-string. Happy to have the f-string fixed in a subsequent PR.

mattwthompson:
Agree with each of you here - that change should work (since the newline is within the f-string magic) but can be done after this PR

Issue #278, opened 2024-04-16
QCSubmit CI is segfaulting


j-wags:
Segfaulting test seems to be `openff/qcsubmit/_tests/test_datasets.py::test_dataset_to_pdf_no_torsions[Openeye by default.]`

https://github.com/openforcefield/openff-qcsubmit/actions/runs/8697856360/job/23853864399#step:7:193

OE hasn't pushed a new package recently ([last main label package was 3 months ago](https://anaconda.org/OpenEye/OpenEye-toolkits/files?channel=main))

mattwthompson:
Having dealt with similar (-ly cryptic) failures in the past, I'd try to parry down which modules that are bundled with OpenEye and might have had releases recently. (I quickly checked a few of the typical suspects and nothing was obvious.)

mattwthompson:
If it helps narrow it down, it's likely in the OEDepict import, so maybe something changed with a visualization-adjacent upstream. (It'd be much messier if this is not related to a visualization tool.)

j-wags:
AFAICT the openeye conda packages vendor everything. They only have python and pytest as deps. Am I overlooking a source of outside code?

<img width="1122" alt="Screenshot 2024-04-16 at 11 57 38 AM" src="https://github.com/openforcefield/openff-qcsubmit/assets/16329175/e9556bdd-1288-4f3a-850f-bb00e1323f68">

mattwthompson:
Nothing sticks out with naive inspection of the changes in imported Python modules:

```python
In [1]: import sys

In [2]: mod1 = set(sys.modules.keys())

In [3]: from openeye import oechem

In [4]: mod2 = set(sys.modules.keys())

In [5]: from openeye import oedepict

In [6]: mod3 = set(sys.modules.keys())

In [7]: mod1.difference(mod2)
Out[7]: set()

In [8]: mod2.difference(mod1)
Out[8]:
{'_oechem4003',
 'importlib.abc',
 'importlib.resources',
 'importlib.resources._adapters',
 'importlib.resources._common',
 'importlib.resources._legacy',
 'importlib.resources.abc',
 'libs',
 'libs.openeye_platform',
 'oechem',
 'openeye',
 'openeye.oechem',
 'swig_runtime_data4'}

In [9]: mod3.difference(mod1)
Out[9]:
{'_oechem4003',
 '_oedepict2533',
 'importlib.abc',
 'importlib.resources',
 'importlib.resources._adapters',
[...]

PR #279, opened 2024-04-17
Test against toolkit branch

mattwthompson:
https://github.com/openforcefield/openff-toolkit/pull/1862


mattwthompson:
This isn't segfaulting, but it's hanging. I think the toolkit branch resolves #278 but not #275

mattwthompson:
I think this is ultimately duplicated effort

It'd still be nice to get `--durations=20` in

PR #282, opened 2024-06-15
Add `verify` named argument to `to_records`

### LINKED ISSUE #269: Add `verify=False` as an optional named argument to `collection.to_records()`
When a user has to specify `verify=False` to construct a PortalClient, they aren't able to run `to_records` later on, since that makes [a new PortalClient with the default](https://github.com/openforcefield/openff-qcsubmit/blob/a42b582a17b7c38b7acc54e4a7d58d5d3d8586c5/openff/qcsubmit/results/results.py#L384) (`verify=True`)

```python

qc_client = PortalClient("https://api.qcarchive.molssi.org:443")
from openff.qcsubmit.results import (
    BasicResultCollection,
    OptimizationResultCollection,
    TorsionDriveResultCollection,
)

# Pull down the torsion drive records from the 'OpenFF Protein Capped 3-mer Backbones v1.0' dataset.
torsion_drive_result_collection = TorsionDriveResultCollection.from_server(
    client=qc_client,
    datasets="OpenFF Protein Capped 3-mer Backbones v1.0",
    spec_name="default",
)
from openff.qcsubmit.results.filters import SMARTSFilter

filtered_collection = torsion_drive_result_collection.filter(SMARTSFilter(smarts_to_include=["C[SH]"]))
filtered_collection
torsion_drive_records = filtered_collection.to_records()
```

> (returns a "you should set `verify=False`" error)

j-wags:
## Todos

  - [x] Fixes #269
  - [ ] Add tests
  - [ ] Update docstrings
  - [ ] Update releasenotes

## Status
- [ ] Ready to go

### DIFF
--- a/openff/qcsubmit/results/results.py
+++ b/openff/qcsubmit/results/results.py
@@ -189,7 +189,7 @@ class _BaseResultCollection(BaseModel, abc.ABC):
             )
 
     @abc.abstractmethod
-    def to_records(self, verify=True) -> List[Tuple[BaseRecord, Molecule]]:
+    def to_records(self, client_kwargs: Optional[dict] = None) -> List[Tuple[BaseRecord, Molecule]]:
         """Returns the native QCPortal record objects for each of the records referenced
         in this collection along with a corresponding OpenFF molecule object.
         """
@@ -371,7 +371,7 @@ class BasicResultCollection(_BaseResultCollection):
             spec_name,
         )
 
-    def to_records(self, verify=True) -> List[Tuple[SinglepointRecord, Molecule]]:
+    def to_records(self, client_kwargs: Optional[dict] = None) -> List[Tuple[SinglepointRecord, Molecule]]:
         """Returns the native QCPortal record objects for each of the records referenced
         in this collection along with a corresponding OpenFF molecule object.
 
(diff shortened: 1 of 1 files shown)


ntBre:
What do you think about generalizing this to `**client_kwargs` or something similar? In messing with the caching section of the `retrieving_results` example, I think providing the `cache_dir` kwarg to `PortalClient` automates caching throughout qcsubmit. If you could pass this through `to_records` that would speed up those calls too!

I also noticed that `OptimizationResultCollection.to_basic_result_collection`, for example, calls `to_records` internally, so any kwargs (verify or the ** version) probably need to be propagated anywhere `to_records` is called.

ntBre:
I might be getting ahead of myself, but if we decide to support these additional keywords, it might make more sense to have some kind of `default_portal_client` function that we call instead of `PortalClient` directly. That way the user can just override that function, and we don't have to worry about propagating kwargs through a huge fraction of the code base. Something like this?

```python
from openff.qcsubmit import default_portal_client

def my_portal_client(addr):
    return PortalClient(addr, verify=False, cache_dir=".")

default_portal_client = my_portal_client
```

j-wags:
Oh, I really like the `client_kwargs` idea, and good catch noticing `to_records` is called in more places internally. I need to think a little more about how to handle that.

The `default_portal_client` idea goes in the right direction - if we don't want to thread kwargs through everywhere we'll need to modify some sort of global state. But from what I recall, wrangling global state is a big pain. If we go this direction it might be best to offer users something like a context manager to give them a better handle on things.

j-wags:
Superseded by #284

PR #283, opened 2024-06-17
Enable examples CI

### LINKED ISSUE #256: Enable examples CI
Examples are out of date after #195

j-wags:
## Todos
- [x] Fix #256 

## Status
- [x] Ready to go


ntBre:
This commit should get `creating-datasets.ipynb` working, but `retrieving-results.ipynb` is failing at a call to `BasicResultCollection.visualize`, which was removed in #141 without comment as far as I can tell. Should this API be restored? Alternatively, its code could be in-lined into the notebook, or that cell of the notebook could just be deleted.

j-wags:
I'd go for in-lining it in this case. It looks like the heavy lifting was done in `molecules_to_pdf` from `openff.qcsubmit.utils.visualize` which is still in the codebase, so that should work great.

j-wags (commented):
This looks great, approved! (unfortunately I can't click approve on this since it's technically my PR)

Before you merge, could you:
1) Update `docs/releasehistory.md` with this PR (you'll can make a new heading like `## Current development` and subheading `### Examples updated`)
2) (optionally) slice the massive outputs to... not be so massive? Something like changing `print(torsion_drive_result_collection)` to `print(str(torsion_drive_result_collection)[:1000])` through the notebook would be perfect)

j-wags:
Oh, and the socket thing is kinda weird, but it drives an important point home so I don't really mind it. We can improve it in the future if we feel inspired, but this PR is already enough of an improvement to merge.

ntBre:
Awesome, thanks for the review! I updated the release history and added a small `preview_dataset` function to print only the first 5 entries in each dataset. I thought this might look nicer than slicing the output, but it does copy the dataset so I'm happy to go for the fix you proposed if you prefer.

j-wags:
Looks great. It's a little complex but I don't know of an easier way to make it cleaner, and it's way better than it was before. Thanks @ntBre!

j-wags:
This is good to merge whenever you're ready :-)

PR #284, opened 2024-06-21
Enable use of a custom PortalClient

ntBre:
## Description
This PR allows users to supply a custom `PortalClient` constructor to functions that use a client internally (most notably `*ResultCollection.to_records`) by overriding the global `openff.qcsubmit.utils.utils.default_portal_client` function, or through the use of the new `portal_client_manager` context manager. This is especially useful for users needing to supply a keyword like `verify=False` (#282) or for users wanting to take advantage of the caching in QCPortal 0.54.

### Example usage

```python
from openff.qcsubmit.utils.utils import portal_client_manager

with portal_client_manager(lambda addr: PortalClient(addr, cache_dir=".")):
    ...
```

I think an added benefit of this approach is potentially making it easier to mock `PortalClient` for tests. I worked on an example of this for a little while, but you have to implement a little more functionality than I expected initially, at least to call something like `to_records`. Still, I think it's doable and potentially useful.

## Todos
  - [x] Rough draft of interface
  - [ ] Docs
  - [ ] Better test(s)?
  - [ ] Add to examples

## Questions
- [ ] What should the API look like, and where should the code live?

## Status
- [ ] Ready to go


j-wags (commented):
For the test, it'd be good to do the socket trick again, though via a context manager if possible (this could be a handy testing fixture that we could use in other tests as well)

I'd recommend making `default_portal_client` private so that users are nudged toward using the (safer) context manager approach)

ntBre:
Thanks for the review! I've made `default_portal_client` private, added a context manager for the socket trick, and added a second test for the `portal_client_manager` unimaginatively named `test_manager2`. I think this one probably supersedes the old test, so I can delete the original `test_manager` and the dataset I added and rename this one if you also prefer this one.

Should `portal_client_manager` be re-exported in `utils/__init__.py` so that it doesn't have to be imported from `openff.qcsubmit.utils.utils`? I can also move the `no_internet` context manager (and/or rename it), but I wasn't sure if that was better to do now or later when another test needed it.

ntBre:
I added a docstring to the `portal_client_manager`, but I noticed in the docs build that it is not displayed anywhere. This is possibly outside the scope of this PR, but it might be nice to update the doc generation to include this. There is also still a Caching section in the generated docs that points to code that doesn't exist anymore.

Anyway, if the new test is good enough, I think that resolves the Todos besides possibly adding this to the Examples. I included an example usage in the docstring, but I can also add this to an example notebook if you want.

openff/qcsubmit/_tests/utils/test_manager.py
```diff
@@ -0,0 +1,55 @@
...
+        client = PortalClient("https://api.qcarchive.molssi.org:443", cache_dir=d)
+        opt = OptimizationResultCollection.from_server(
+            client,
+            "OpenFF Torsion Multiplicity Optimization Training Coverage Supplement v1.0",
+        )
+        # this fails without the portal_client_manager, as desired. reusing the
+        # same client in the lambda prevents the network access that occurs
+        # when creating the client
+        with portal_client_manager(lambda _: client), no_internet():
+            opt.to_records()
```
j-wags, line 55: (blocking) Since we're doing spooky global python stuff, could you add a check that, after the context manager closes, things behave as predicted? I'd suggest something roughly like:

```suggestion
        # this fails without the portal_client_manager, as desired. reusing the
        # same client in the lambda prevents the network access that occurs
        # when creating the client
        with portal_client_manager(lambda _: client), no_internet():
            opt.to_records()
        with no_internet():
            with pytest.mark.raises(Exception, match="socket accessed"):
                opt.to_records()
```

openff/qcsubmit/_tests/utils/test_manager.py
```diff
@@ -0,0 +1,55 @@
...
+    """Retrieve a full dataset from QCArchive to populate the cache, then
+    access the cache offline in `to_records` by passing the same client args to
+    the `portal_client_manager`
+    """
+    with TemporaryDirectory() as d:
+        client = PortalClient("https://api.qcarchive.molssi.org:443", cache_dir=d)
+        opt = OptimizationResultCollection.from_server(
+            client,
+            "OpenFF Torsion Multiplicity Optimization Training Coverage Supplement v1.0",
+        )
```
j-wags, line 50: (not blocking) It might be worth doing the `assert (Path(d) / "api.qcarchive.molssi.org_443").exists()` thing in this test as well.

openff/qcsubmit/_tests/utils/test_manager.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def test_manager():
+    ds = OptimizationResultCollection.parse_file(get_data("tiny-opt-dataset.json"))
+    with (
+        TemporaryDirectory() as d,
+        portal_client_manager(lambda addr: PortalClient(addr, cache_dir=d)),
+    ):
+        ds.to_records()
+        assert (Path(d) / "api.qcarchive.molssi.org_443").exists()
```
j-wags, line 37: (not blocking) I think `test_manager2` is a strict improvement on this test, so this (and `tiny-opt-dataset.json`) can be removed.

j-wags (changes requested):
Thanks, I think this has the big pieces in place and the docstring is wonderfully thorough. There are a few changes I'd like to see - The ones that will block approval begin with (blocking) while the ones that I think would be good but won't block merging if you disagree are labeled (not blocking) 

Also, (not blocking) could you try adding `utils.py` to the API docs? I can't explain why the current docs have [a `Utils` section that doesn't include anything from `utils.py`](https://github.com/openforcefield/openff-qcsubmit/blame/main/docs/api.rst#L185), so we could rename the old section to `Workflow Utilities` and the new one `Utilities`. (If you have trouble getting this to work, we could consult with docs guru @Yoshanuikabundi when he comes back from vacation next week)

ntBre:
I think I handled all of the reviews! I also rebased onto main and force pushed here, not realizing that that might complicate the reviews. Sorry in advance if that's the case.

I think the only change I made outside of your suggestions was re-exporting the `portal_client_manager` from `utils.utils` to just `utils`. Hopefully that's okay. I also went ahead and updated the release notes so that this will be ready to merge when all the reviews are resolved.

I think the docs look pretty good, but from a quick search, I think I might need double backquotes to get the inline code to render as I expected instead of the single backquotes I used. The parameter section also seems a little off. I'm happy to tweak both of those if you want, but I figured I would comment here while the checks have passed before I start pushing more docstring changes.

image

ntBre:
I fixed the issues I saw in the docs, should be ready for another review!

PR #286, opened 2024-06-26
Use `get_records_with_cache` to cache `to_records` calls

ntBre:
## Description
This PR uses the [get_records_with_cache](https://github.com/MolSSI/QCFractal/blob/5e35393e150fb67a528f26549b6f9ea3bc68c186/qcportal/qcportal/cache.py#L590) function discussed in our last QCArchive meeting to cache `to_records` calls automatically. With these changes alone, users cannot actually access this behavior, but when combined with #284, it enables code like this:

```python

from openff.qcsubmit._tests.utils.test_manager import no_internet
from openff.qcsubmit.results import OptimizationResultCollection
from openff.qcsubmit.utils.utils import portal_client_manager

opt = OptimizationResultCollection.parse_file("tiny-opt-dataset.json")
client = PortalClient("https://api.qcarchive.molssi.org:443", cache_dir=".")

with portal_client_manager(lambda _: client):
    print(len(opt.to_records()))
    with no_internet():
        print(len(opt.to_records()))
```

For this PR, I applied the changes separately from #284 to the main branch, but I have used this code to test the combined changes locally, and it works! I have not run into MolSSI/QCFractal#844 here yet, so hopefully that is less common in real use cases.

## Todos
  - [x] Enable caching for expensive `to_records` calls
  - [x] Probably wait for #284. This isn't really useful on its own, but I was excited to get a proof of concept working, and I think both changes are actually needed for either one to help my valence fits

## Status
- [ ] Ready to go

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -230,6 +230,7 @@ Utilities
     :nosignatures:
     :toctree: api/generated/
 
+	CachedPortalClient
 	portal_client_manager
 
 Exceptions
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -230,7 +230,7 @@ Utilities
     :nosignatures:
     :toctree: api/generated/
 
-	CachedPortalClient
+	_CachedPortalClient
 	portal_client_manager
 
 Exceptions
--- a/openff/qcsubmit/utils/utils.py
+++ b/openff/qcsubmit/utils/utils.py
@@ -43,6 +43,7 @@ class CachedPortalClient(PortalClient):
         *,
         include: Optional[Iterable[str]] = None,
     ) -> Union[Optional[OptimizationRecord], List[Optional[OptimizationRecord]]]:
+        raise NotImplementedError()
         if missing_ok:
             logger.warning("missing_ok provided but unused by CachedPortalClient")
         if unpack := not isinstance(record_ids, Sequence):
@@ -67,6 +68,7 @@ class CachedPortalClient(PortalClient):
         *,
         include: Optional[Iterable[str]] = None,
     ) -> Union[Optional[SinglepointRecord], List[Optional[SinglepointRecord]]]:
+        raise NotImplementedError()
         if missing_ok:
             logger.warning("missing_ok provided but unused by CachedPortalClient")
         if unpack := not isinstance(record_ids, Sequence):
@@ -91,6 +93,7 @@ class CachedPortalClient(PortalClient):
         *,
         include: Optional[Iterable[str]] = None,
     ) -> Union[Optional[TorsiondriveRecord], List[Optional[TorsiondriveRecord]]]:
+        raise NotImplementedError()
         if missing_ok:
             logger.warning("missing_ok provided but unused by CachedPortalClient")
         if unpack := not isinstance(record_ids, Sequence):
--- a/openff/qcsubmit/utils/utils.py
+++ b/openff/qcsubmit/utils/utils.py
@@ -43,7 +43,6 @@ class CachedPortalClient(PortalClient):
         *,
         include: Optional[Iterable[str]] = None,
     ) -> Union[Optional[OptimizationRecord], List[Optional[OptimizationRecord]]]:
-        raise NotImplementedError()
         if missing_ok:
             logger.warning("missing_ok provided but unused by CachedPortalClient")
         if unpack := not isinstance(record_ids, Sequence):
--- a/openff/qcsubmit/utils/utils.py
+++ b/openff/qcsubmit/utils/utils.py
@@ -26,6 +26,8 @@ from qcportal.torsiondrive.record_models import TorsiondriveRecord
 
 logger = logging.getLogger(__name__)
 
+_DEFAULT_CACHE_DIR = "./qcsubmit_qcportal_cache"
+
 
 class CachedPortalClient(PortalClient):
     def __init__(self, addr, cache_dir, **client_kwargs):
@@ -108,7 +110,7 @@ class CachedPortalClient(PortalClient):
 
 
 def _default_portal_client(client_address) -> PortalClient:
-    return CachedPortalClient(client_address, cache_dir="./qcsubmit_qcportal_cache")
+    return CachedPortalClient(client_address, cache_dir=_DEFAULT_CACHE_DIR)
 
 
 @contextmanager
--- a/openff/qcsubmit/utils/utils.py
+++ b/openff/qcsubmit/utils/utils.py
@@ -1,5 +1,17 @@
+import logging
+import os
 from contextlib import contextmanager
-from typing import Callable, Dict, Generator, List, Tuple
+from typing import (
+    Callable,
+    Dict,
+    Generator,
+    Iterable,
+    List,
+    Optional,
+    Sequence,
+    Tuple,
+    Union,
+)
 
 from openff.toolkit import topology as off
 from openff.toolkit.utils.toolkits import (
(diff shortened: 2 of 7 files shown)


ntBre:
After thinking more about this, possibly a better solution would be for us (or actually a user) to subclass `PortalClient` and override the various `get_{singlepoints,optimizations,torsiondrives}` methods with cached versions. That would avoid having to add the `record_cache` field to the standard `PortalClient`, which felt weird to me, and revert all of these changes because they would be handled by providing this new `CachedPortalClient`. I'll try this approach tomorrow.

ntBre:
Here's an example of the `PortalClient` subclass. Unfortunately, I also found out that the `no_internet` trick is not infallible. I think the fact that the `PortalClient` opens a `requests.Session` and holds onto it means that it can access the internet without going back through `socket.socket`. However, I monitored internet traffic with Wireshark and made sure that this code does not access the internet on the second `to_records` call, which is also verified by setting `_req_session` to `None`.

```python
import os
import shutil

from qcportal.cache import RecordCache, get_records_with_cache
from qcportal.optimization import OptimizationRecord

from openff.qcsubmit._tests.utils.test_manager import no_internet
from openff.qcsubmit.results import OptimizationResultCollection
from openff.qcsubmit.utils.utils import portal_client_manager


class CachedPortalClient(PortalClient):
    def __init__(self, addr, cache_dir, **client_kwargs):
        super().__init__(addr, cache_dir=cache_dir, **client_kwargs)
        self.record_cache = RecordCache(
            os.path.join(self.cache.cache_dir, "cache.sqlite"), read_only=False
        )

    def get_optimizations(self, record_ids, missing_ok=False, include=None):
        return get_records_with_cache(
            self,
            self.record_cache,
            OptimizationRecord,
            record_ids,
            include=["initial_molecule", "final_molecule"],
        )


if os.path.exists("api.qcarchive.molssi.org_443"):
    shutil.rmtree("api.qcarchive.molssi.org_443")

opt = OptimizationResultCollection.parse_file("tiny-opt-dataset.json")
client = CachedPortalClient("https://api.qcarchive.molssi.org:443", cache_dir=".")

with portal_client_manager(lambda _: client):
    client._req_session = None
    with no_internet():
        print(len(opt.to_records()))
```

This code runs on the branch for #284 without any changes to qcsubmit. The implementation of `CachedPortalClient` basically needs to know which methods we call on `PortalClient` internally, though, so I think it might make sense for us to provide this class even though a user could define it.

ntBre:
I've updated the implementation to use the `CachedPortalClient` from the code block above. We discussed making the class private, but I left it public for now because I think it might be useful for anyone wanting to use additional `PortalClient` kwargs with the `portal_client_manager`. I'm happy to make it private if you prefer, though. That's a pretty advanced use case and maybe not one we want to support in the public API.

I have not added any tests specifically for the caching yet, but I think it's a great first sign that the tests pass after replacing the default `PortalClient` with this cached version. I also tried intentionally introducing errors in each of the new methods to make sure they were actually being called, so they are all covered by existing tests at least.

One issue I've run into with testing this is that each test uses the same default `cache_dir`. I tried adding a pytest fixture to delete the cache dir before each test ran, but this caused disastrous issues with pytest-xdist running multiple tests in parallel (one would be trying to write to the cache while the next test tried deleting the dir). I think this means that fully testing the caches will be more involved than I hoped. It also means that I currently have to delete the cache dir locally between each test run, or the tests just use the cache. That won't be an issue in CI at least. My best idea for fixing this currently is to modify the tests using `PortalClient`s more invasively to wrap the code in `portal_client_managers` with temporary cache dirs so that each test can run fully independently and clean itself up afterward.

ntBre:
The tests should now be passing with temporary cache directories everywhere `CachedPortalClient` is used. In my local tests I was seeing ignored instances of the error from MolSSI/QCFractal#844, but it sounds like that has been fixed in the next version (0.56) of qcportal.

Now, I think I just need to update these tests to check that the caching actually worked.

ntBre:
The tests have been updated, and I copied over (and modified) documentation from qcportal. It should be ready for review!

openff/qcsubmit/utils/utils.py
```diff
@@ -1,16 +1,256 @@
...
+            Additional fields to include in the returned record
+
+        Returns
+        -------
+        :
+            If a single ID was specified, returns just that record. Otherwise,
+            returns a list of records.
+        """
+        if missing_ok:
+            logger.warning("missing_ok provided but unused by CachedPortalClient")
```
j-wags, line 173: (not blocking) ditto

openff/qcsubmit/utils/utils.py
```diff
@@ -1,16 +1,256 @@
...
```
j-wags, line 219: (not blocking) ditto

docs/api.rst
```diff
@@ -230,6 +230,7 @@ Utilities
     :nosignatures:
     :toctree: api/generated/
 
+	CachedPortalClient
```
j-wags, line 233:
```suggestion
	_CachedPortalClient
```

openff/qcsubmit/utils/utils.py
```diff
@@ -1,16 +1,256 @@
...
+            include=include,
+            force_fetch=False,
+        )
+        if unpack:
+            return res[0]
+        else:
+            return res
+
+    @contextmanager
+    def _no_session(self):
```
j-wags, line 236: (not blocking) this looks like a good addition. If this is strictly better than `no_internet`, then I'd be in favor of replacing all usages of the latter in a subsequent PR.

j-wags (approved):
Thanks @ntBre. We reviewed this during our check in this week and I summarized our discussion in the comments. This is good to merge once the (blocking) items are resolved and the releasenotes are updated!

openff/qcsubmit/utils/utils.py
```diff
@@ -1,16 +1,256 @@
...
+        self._req_session = None
+        try:
+            yield
+        finally:
+            self._req_session = tmp
 
 
 def _default_portal_client(client_address) -> PortalClient:
-    return PortalClient(client_address)
+    return CachedPortalClient(client_address, cache_dir=_DEFAULT_CACHE_DIR)
```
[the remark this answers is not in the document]
ntBre: Good call on this. This code fails with a `database is locked` error:

```python
from multiprocessing import Pool

from openff.qcsubmit.results import TorsionDriveResultCollection
from openff.qcsubmit.utils import CachedPortalClient, portal_client_manager

ds = TorsionDriveResultCollection.parse_file(
    "/home/brent/omsf/projects/valence-fitting/02_curate-data/datasets/supp-td.json"
)
datasets = [ds, ds, ds, ds]
client = CachedPortalClient("https://api.qcarchive.molssi.org:443", cache_dir=".")
with portal_client_manager(lambda _: client), Pool(4) as pool:
    for res in pool.imap_unordered(
        TorsionDriveResultCollection.to_records, datasets, 1
    ):
        print(len(res))
```

It also fails without the `portal_client_manager` and `CachedPortalClient` as the default client, so it's not safe to share the same client across threads or to construct multiple clients accessing the same database.

It actually doesn't quite work most of the time, even with a regular `PortalClient` with or without a cache dir, but some of the processes have `read` timeout errors rather than the more dramatic database errors from the cached version. It's probably best in general not to share a `PortalClient` across threads.

Using `with portal_client_manager(PortalClient), Pool(4) as pool:` (removing the lambda that shares the same client), as your suggestion will restore the default to, has worked successfully on every run I've tried so far, so I think that's definitely the right default.

Do you think it's worth mentioning something about this in the documentation for `portal_client_manager`? The docs currently show an example of a function that returns a new `PortalClient` each time (which should be safe) rather than a `lambda _: client`. I'm not really sure how to warn against this without showing the lambda code in the docs, though.

openff/qcsubmit/utils/utils.py
ntBre: I think you're right about this too. I didn't know it at the time, but `test_manager` (the only usage of `no_internet`) only looks like it works because it's reusing the client with the lambda (thus preserving the session), not because it's reading from the cache. In fact, setting `client._req_session = None` in that test causes it to fail because the dataset fetching cache and the cache needed by the `to_records` call are separate. Now that `portal_client_manager` is used in many other tests, I think it would be okay to delete the `test_manager.py` file with its single test and the `no_internet` manager entirely. 

Do you want me to do that here, or in a separate PR as you mentioned?

docs/api.rst
```diff
@@ -262,4 +263,4 @@ Exceptions
     PCMSettingError
     InvalidDatasetError
     DatasetRegisterError
-    RecordTypeError
\ No newline at end of file
+    RecordTypeError
```
ntBre, line 266: Emacs kept trying to add this newline to the end of the file each time I edited it. I finally forgot to revert the change before committing, hopefully that's okay.

openff/qcsubmit/utils/utils.py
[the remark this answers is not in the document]
j-wags: Thanks for the update and for checking on multithreading. I've read this a few times and think I kinda understand, and my preference is to do the simplest safe thing. So I agree with reverting the default behavior. And updating the docs is optional, you might just copy your comment "It's not safe to share the same client across threads or to construct multiple clients accessing the same database" into the portal_client_manager docstring and call it a day.

openff/qcsubmit/utils/utils.py
j-wags: Good catch about the test not doing what we thought. Let's make those changes in a separate PR - This one has so much scope that it's getting tough to review, and it'll be good to consolidate our gains.

PR #290, opened 2024-07-11
DDX solvent support 

jthorton:
## Description
This PR adds support for the new DDX solvent model interface in Psi4 which should be slightly faster than PCM and scale better with molecule size. The other benefit is direct control over the solvent dielectric constant, which will be used in an upcoming charge model dataset. 

## Todos
  - [x] add ddx option validation
  - [x] add tests

## Questions
- [ ] Question1

## Status
- [x] Ready to go

### DIFF
--- a/openff/qcsubmit/_tests/test_submissions.py
+++ b/openff/qcsubmit/_tests/test_submissions.py
@@ -12,7 +12,7 @@ from qcportal import PortalClient
 from qcportal.record_models import RecordStatusEnum
 
 from openff.qcsubmit import workflow_components
-from openff.qcsubmit.common_structures import MoleculeAttributes, PCMSettings
+from openff.qcsubmit.common_structures import MoleculeAttributes, PCMSettings, DDXSettings
 from openff.qcsubmit.constraints import Constraints
 from openff.qcsubmit.datasets import (
     BasicDataset,
@@ -281,7 +281,11 @@ def test_basic_submissions_multiple_spec(fulltest_client):
             assert record.specification == spec
 
 
-def test_basic_submissions_single_pcm_spec(fulltest_client):
+@pytest.mark.parametrize("solvent_model, solvent_energy", [
+    pytest.param(PCMSettings(units="au", medium_Solvent="water"), "pcm polarization energy", id="PCM"),
+    pytest.param(DDXSettings(ddx_solvent_epsilon=4), "dd solvation energy", id="DDX")
+])
+def test_basic_submissions_single_solvent_spec(fulltest_client, solvent_model, solvent_energy):
     """Test submitting a basic dataset to a snowflake server with pcm water in the specification."""
 
     client = fulltest_client
@@ -299,7 +303,7 @@ def test_basic_submissions_single_pcm_spec(fulltest_client):
         program=program,
         spec_name="default",
         spec_description="testing the single points with pcm",
-        implicit_solvent=PCMSettings(units="au", medium_Solvent="water"),
+        implicit_solvent=solvent_model,
         overwrite=True,
     )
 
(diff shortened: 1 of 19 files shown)


openff/qcsubmit/_tests/test_submissions.py
```diff
@@ -336,17 +359,23 @@ def test_basic_submissions_single_pcm_spec(fulltest_client):
...
             assert record.return_result is not None
             # make sure the PCM result was captured
-            assert record.extras["qcvars"]["PCM POLARIZATION ENERGY"] < 0
-            assert record.specification == spec
+            assert record.properties[solvent_energy] < 0
+            # make sure the correct solvent was used
+            assert solvent_evidence in record.stdout
+            assert record.specification.dict(
+                include={"method", "basis", "program"}
+            ) == spec.dict(include={"method", "basis", "program"})
```
ntBre, line 378: (non-blocking) I see that this used to be a direct equivalence check. Are there fields in the specs that differ now?

openff/qcsubmit/common_structures.py
```diff
@@ -400,6 +415,44 @@ def to_string(self) -> str:
...
+    """
+    A simple settings class for the ddx solvent model to be used with Psi4
+    """
+
+    ddx_model: Literal["pcm", "cosmo"] = Field(
+        "pcm", description="The solvation model to use."
+    )
+    ddx_radii_scaling: PositiveFloat = Field(
+        1.1,
+        description="The scaling factor for the cavity spheres this also depends on the radii set chosen.",
```
ntBre, line 434: not blocking, very minor
```suggestion
        description="The scaling factor for the cavity spheres. This also depends on the radii set chosen.",
```

openff/qcsubmit/common_structures.py
```diff
@@ -400,6 +415,44 @@ def to_string(self) -> str:
...
+    ddx_radii_set: Literal["uff", "bondi"] = Field(
+        "uff", description="The atomic radii set to use."
+    )
+    ddx_solvent_epsilon: Optional[float] = Field(
+        None,
+        description="The dielectric constant of the solvent, if not specified the solvent type should be set.",
+    )
+    ddx_solvent: str = Field(
+        "water",
+        description="The name of the ddx supported solvent which should used, the epsilon value will be determined from `pyddx.data.solvent_epsilon`. Note that this value is ignored if the `ddx_solvent_epsilon` is provided.",
```
ntBre, line 445: not blocking, very minor
```suggestion
        description="The name of the ddx supported solvent which should be used, the epsilon value will be determined from `pyddx.data.solvent_epsilon`. Note that this value is ignored if the `ddx_solvent_epsilon` is provided.",
```

openff/qcsubmit/common_structures.py
```diff
@@ -400,6 +415,44 @@ def to_string(self) -> str:
...
+    )
+
+    def add_keywords(self, keyword_data: dict) -> dict:
+        keyword_data["ddx"] = True
+        ddx_data = self.dict()
+        if self.ddx_solvent_epsilon is None:
+            del ddx_data["ddx_solvent_epsilon"]
+        keyword_data.update(self.dict())
```
ntBre, line 453: (blocking) This should be `update(ddx_data)` right? Otherwise the `del` above won't apply? If that's the case, it might be nice to throw in a test case where `ddx_solvent_epsilon` is unset to make sure too.

ntBre (approved):
Sorry for the delay on the review, this looks great overall! I'm approving, but please take a look at the comments before merging. In particular, the one I marked (blocking) is a possible correctness issue, if I understand how the current code is working. I'd also prefer if the test were still a simple `==` check, but I marked it non-blocking since I assume you changed it for a reason. The other two suggestions are minor wording changes that you can ignore if you want.

ntBre:
Oh, please also update the release notes!

PR #291, opened 2024-07-22
Re-enable codecov

ntBre:
Based on the CI errors, I think we're just missing the token. I also changed `fail_ci_if_error` to `true` in-line with the toolkit repo.

## Description
Enable codecov coverage checks in CI. There was already a step in CI.yml for this, but no token was being provided and `fail_ci_if_error` was `false`, so it was silently failing each time.

## Status
- [ ] Ready to go

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -83,5 +83,6 @@ jobs:
       - name: Codecov
         uses: codecov/codecov-action@v4
         with:
+          token: ${{ secrets.CODECOV_TOKEN }}
           file: ./coverage.xml
-          fail_ci_if_error: false
+          fail_ci_if_error: true


mattwthompson:
I added a token from [^1] to [^2] and re-kicked CI for fun


[^1]: https://app.codecov.io/gh/openforcefield/openff-qcsubmit/settings

[^2]: https://github.com/openforcefield/openff-qcsubmit/settings/secrets/actions

.github/workflows/CI.yml
```diff
@@ -83,5 +83,6 @@ jobs:
       - name: Codecov
         uses: codecov/codecov-action@v4
         with:
+          token: ${{ secrets.CODECOV_TOKEN }}
           file: ./coverage.xml
-          fail_ci_if_error: false
+          fail_ci_if_error: true
```
mattwthompson, line 88: IMHO this should be the default, but Sentry disagrees 🤷

mattwthompson (approved):
LGTM % re-kicking the run that failed because Psi4 didn't finish in time

.github/workflows/CI.yml
ntBre: Agreed

j-wags:
Wow - opened and resolved in the time it took me to meet a collaborator for lunch. @mattwthompson's approval is all that's needed here. Thanks @ntBre!

PR #295, opened 2024-09-23
Use new Pydantic v1 backdoor

mattwthompson:
## Description

https://docs.pydantic.dev/latest/changelog/#v11018-2024-08-22), 

## Todos
  - [x] Use new Pydantic v1 backdoor
  - [x] Consolidate some duplicate code

## Status
- [ ] Ready to go

### DIFF
--- a/devtools/conda-envs/basic.yaml
+++ b/devtools/conda-envs/basic.yaml
@@ -39,7 +39,7 @@ dependencies:
 
   - openff-toolkit
   - openff-units >=0.2.1
-  - pydantic >1.10.18,<3
+  - pydantic >1.10.17,<3
   - pyyaml
   - torsiondrive
   - basis_set_exchange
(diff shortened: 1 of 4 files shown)


mattwthompson:
Dozens of 

```
FAILED openff/qcsubmit/_tests/test_datasets.py::test_dataset_update[Dataset no metadata] - AttributeError: 'apsw.Connection' object has no attribute 'pragma'

File: devtools/conda-envs/basic.yaml [hunk 1]
```diff
@@ -39,7 +39,7 @@ dependencies:
 
   - openff-toolkit
   - openff-units >=0.2.1
-  - pydantic
+  - pydantic >1.10.17,<3
```
ntBre, line 42: This is probably my packaging naivete showing, but is there a reason not to do `>=2`? I see a version 1.10.18 in the pydantic docs, which would be allowed by these constraints right? Maybe it also has the v1 module.

Issue #299, opened 2024-10-03
BasicResultCollection.from_server crashes if an entry is missing a CMILES


amcisaac:
When downloading a `BasicResultCollection` using `BasicResultCollection.from_server()`, it crashes if it encounters an entry that is missing a CMILES, with the error:

```
Traceback (most recent call last):
  File "/dfs6/pub/amcisaac/sage-2.2.1-sdata/03_generate-initial-ff/create-msm-ff.py", line 284, in <module>
    main()
  File "/data/homezvol3/amcisaac/miniconda3/envs/sep-2024-env/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/homezvol3/amcisaac/miniconda3/envs/sep-2024-env/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/data/homezvol3/amcisaac/miniconda3/envs/sep-2024-env/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/homezvol3/amcisaac/miniconda3/envs/sep-2024-env/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/dfs6/pub/amcisaac/sage-2.2.1-sdata/03_generate-initial-ff/create-msm-ff.py", line 133, in main
    hess_dataset = BasicResultCollection.from_server(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/homezvol3/amcisaac/miniconda3/envs/sep-2024-env/lib/python3.11/site-packages/openff/qcsubmit/results/results.py", line 365, in from_server
    return cls.from_datasets(
           ^^^^^^^^^^^^^^^^^^
  File "/data/homezvol3/amcisaac/miniconda3/envs/sep-2024-env/lib/python3.11/site-packages/openff/qcsubmit/results/results.py", line 324, in from_datasets
    cmiles = entry.attributes[
             ^^^^^^^^^^^^^^^^^
[...]

PR #300, opened 2024-10-04
Use `get` to avoid KeyError on missing cmiles attribute

ntBre:
## Description
This fixes the crash described in #299, although it doesn't add any additional checks to warn if every entry was skipped.

## Todos
  - [x] Fixes crash in #299
  - [x] Fixes related potential issue for `OptimizationResultCollection`

## Questions
- [ ] Should a similar check be performed in `TorsionDriveResultCollection`? Currently it only tries `entry.attributes` for the CMILES and will crash if it's not found.

## Status
- [ ] Ready to go


ntBre:
Thanks for the review! I think I covered everything we discussed.

ntBre:
I also quoted this in one of the commit messages, but I'll copy it here as a reason for switching to `logging` from `warnings`. In practical terms I just didn't see an easy way to filter a specific warning class by default as we discussed in the meeting yesterday, but I thought the docs were an interesting additional justification.

> based on these docs (https://docs.python.org/3/howto/logging.html#when-to-use-logging) warnings
> should be used "if the issue is avoidable and the client application should be
> modified to eliminate the warning," (like a deprecation warning) while logging a
> warning should be used "if there is nothing the client application can do about
> the situation, but the event should still be noted." I think this warning is
> closer to the latter case, where there's not much a user can do about this, and
> this allows us to filter these warnings by default, which I don't think is
> possible with the warnings module

j-wags (approved):
Awesome, thanks for the revisions and for looking into logging vs. warnings best practices @ntBre! 

Please update the release notes before you merge :-)

Issue #297, opened 2024-10-02
Not all records being found for Hessian Dataset when converting from Optimization Dataset


amcisaac:
I have an [optimization](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2024-09-11-OpenFF-Sulfur-Optimization-Training-Coverage-Supplement-v1.0) dataset, and a corresponding [single point dataset](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2024-09-18-OpenFF-Sulfur-Hessian-Training-Coverage-Supplement-v1.0) with Hessians for all the final molecules. Typically we fetch the hessian records using `dataset.to_basic_result_collection(driver="hessian")`, but this is not fetching all the Hessian records, even though they should exist.

Reproducing example:

```
from qcportal import PortalClient
from openff.qcsubmit.results import OptimizationResultCollection
from openff.qcsubmit.results import BasicResultCollection

client = PortalClient(address="https://api.qcarchive.molssi.org:443/")

opt_set = OptimizationResultCollection.from_server(client = client, datasets = ['OpenFF Sulfur Optimization Training Coverage Supplement v1.0'],spec_name='default')

opt_to_hess_set = opt_set.to_basic_result_collection(driver="hessian")

print(opt_to_hess_set.n_results) # 260

print(opt_set.n_results) # 899

hess_set = BasicResultCollection.from_server(client=client, datasets=['OpenFF Sulfur Hessian Training Coverage Supplement v1.0'],spec_name='default')

print(hess_set.n_results) # 899

# Check whether all the Basic Results actually have Hessians
hess_rec = hess_set.to_records()

n_hess = 0
for record, molecule in hess_rec:
    if len(record.return_result)>1:
        n_hess += 1
    

print(n_hess) # 899

n_hess = 0
for record, molecule in hess_rec:
    if record.specification.driver == 'hessian':
        n_hess += 1
    

print(n_hess) # 899
```

Expected behavior:
```
test_opt = OptimizationResultCollection.from_server(client=client,datasets = ['OpenFF Gen 2 Opt Set 3 Pfizer Discrepancy'],spec_name='default')

test_opt_to_hess = test_opt.to_basic_result_collection(driver="hessian")

print(test_opt.n_results) # 197

test_opt_to_hess.n_results # 197
```

amcisaac:
cc @ntBre

ntBre:
[...]

PR #303, opened 2024-10-15
Update `OptimizationResultCollection.create_basic_dataset` to preserve molecule IDs

ntBre:
## Description
The goal of this PR is a long-term fix to #297. In short, the current `OptimizationResultCollection.create_basic_dataset` implementation creates a new dataset by round-tripping through OpenFF `Molecule`s, which causes issues with the hashing/equivalence checks in QCArchive, leading to "the same molecules" being considered different. In turn, this causes `OptimizationResultCollection.to_basic_result_collection` to return fewer records than expected.

Earlier qca-dataset-submissions ([example](https://github.com/openforcefield/qca-dataset-submission/blob/95809b4d5a98a118e340b393fdd3636ad24a8606/submissions/2019-07-05%20OpenFF%20NCI250K%20Boron%201/04_create_hessian_dataset.py#L4)) manually constructed qcportal datasets instead of using qcsubmit datasets to avoid this round trip. Due to the major updates to qcportal, I don't think this code works anymore, and it seems much more ergonomic to handle this in create_basic_dataset anyway.

The fix is simply to pass qcelemental `Molecule` objects along directly instead of reconstructing OpenFF `Molecule`s. As it turns out, this was essentially already built into qcsubmit and just required a two-line change to update some keyword arguments in the `BasicDataset.add_molecule` call.

## Todos

  - [x] Update `create_basic_dataset` to preserve molecule hashes from the optimization dataset

## Status
- [x] Ready to go

### DIFF
--- a/openff/qcsubmit/_tests/results/test_results.py
+++ b/openff/qcsubmit/_tests/results/test_results.py
@@ -382,8 +382,9 @@ def test_optimization_create_basic_dataset(optimization_result_collection):
 
 
 def test_optimization_create_basic_dataset_297(public_client):
-    """
-    Test creating a new ``BasicDataset`` from the result of an optimization dataset.
+    """Test creating a new ``BasicDataset`` from the result of an optimization
+    dataset, and verify that the molecule hashes match to prevent the creation
+    of separate records on QCArchive. See issue #297 for more details.
     """
 
     opt = OptimizationResultCollection.from_server(
--- a/openff/qcsubmit/_tests/results/test_results.py
+++ b/openff/qcsubmit/_tests/results/test_results.py
@@ -381,16 +381,39 @@ def test_optimization_create_basic_dataset(optimization_result_collection):
     assert dataset.n_records == 5  # the collection contains 1 duplicate
 
 
-def test_optimization_create_basic_dataset_297(public_client):
+def test_optimization_create_basic_dataset_297():
     """Test creating a new ``BasicDataset`` from the result of an optimization
     dataset, and verify that the molecule hashes match to prevent the creation
     of separate records on QCArchive. See issue #297 for more details.
     """
 
-    opt = OptimizationResultCollection.from_server(
-        public_client,
-        datasets=["OpenFF Sulfur Optimization Training Coverage Supplement v1.0"],
-        spec_name="default",
+    # these are two real entries from the "OpenFF Sulfur Optimization Training
+    # Coverage Supplement v1.0" dataset used in issue #297. the first fails the
+    # round-trip in the previous create_basic_dataset implemenation but the
+    # second works with either
+    opt = OptimizationResultCollection.parse_raw(
+        """
+        {
+            "entries": {
+                "https://api.qcarchive.molssi.org:443/": [
+                {
+                    "type": "optimization",
+                    "record_id": 138340682,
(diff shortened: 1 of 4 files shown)


ntBre:
Well, that turned out to be a pretty easy fix. The hardest part was adding a new mock `get_molecules` method to the `_PortalClient` test fixture. The main limitation now is just that the new test is pulling down a full dataset from a live QCArchive instance. I think I can probably pull out a small subset of the real dataset to speed that up at least.

openff/qcsubmit/_tests/results/test_results.py
```diff
@@ -381,6 +381,59 @@ def test_optimization_create_basic_dataset(optimization_result_collection):
...
+    opt_hashes = {rec.final_molecule.get_hash() for rec, _mol in opt.to_records()}
+
+    basic = opt.create_basic_dataset(
+        "dummy basic dataset name",
+        "descdesc",
+        "tagtagtag",
+        driver="hessian",
+    )
+
+    bas_hashes = {mol.molecule.get_hash() for mol in basic._get_entries()}
```
j-wags, line 428: (blocking) I recall that about 1/3rd of the original dataset still had an identical hash, even after whatever numerical mangling had happened. Given that those molecules had tens of atoms with coordinates, any one of which could have triggered the coordinate rounding error, and it still only happened 1/3rd of the time, this should use an input that we know will trigger the error instead of relying on a molecule as small as methane (with only 5 coords) and the rdkit-backed `offmol.generate_conformers` method (which could change behavior when using a different version of rdkit).

j-wags (changes requested):
Thanks @ntBre , I wrapped my head around the issue and I think this is a good fix. However I think the test could be improved - I'm operating on the assumption that some sort of mangling/precision error with the coordinates was the ultimate cause of the hash mismatch, and I'm not convinced that the current test would have failed with the previous logic. Can you think of a way to make it better replicate the real issue?

ntBre:
Hey Jeff, I'll take a closer look next week, but this test does fail with the previous logic. I wrote it first and tested that in my first commit.

I also minimized the real dataset including one of the molecules that failed in Lexie's report so I'm not quite following the methane comment.

Anyway, thanks again, and I'll take another look on Monday.

j-wags:
(happy to pick this up at our check in monday, have a good weekend!) Oh, maybe I'm misunderstanding. I was under the impression that both molecules only had the coordinates from methane (I didn't think they were fetching their full coords from QCArchive because of how `PortalClient.get_molecules` is overridden). Is it instead the case that methane there is a placeholder and that the full coordinates are being fetched before the hash comparisons?

ntBre:
Ah I see now, I forgot I used methane in the mock. I probably should have emphasized this more in the PR itself, but the `get_molecules` mock is not for the new test I wrote, which uses a live PortalClient from qcportal. The mock is only used by the unchanged, existing test right above the new one that uses the mock _PortalClient class.

I think the coverage of the new test should overlap with the old one, so it might be less confusing just to delete the old test and the new mock. But that seems good to discuss at our meeting!

j-wags (commented):
From our one-on-one today - We really don't want this hacky `get_molecules` mock, so @ntBre will make the new test fully supersede the test above (`test_optimization_create_basic_dataset`), which is almost done but will additionally require finding two records with identical CMILES but different coordinates to run the `n_molecules` vs. `n_records` assertions.

openff/qcsubmit/_tests/results/test_results.py
j-wags: Resolving because this wasn't using methane coords at all, THIS test was actually reaching out to real QCA to get real confs (and in particular, ones that BW established DO have the bad behavior in question)

PR #307, opened 2024-11-26
Use `ruff` for formatting

mattwthompson:
## Description
It's the thing

## Status
- [ ] Ready to go


mattwthompson:
The failures are due to overriding defaults (Pydantic's `.dict()` method and some classes' `type` attribute). I'll get to this but not soon, since the benefit is mostly to development speed

mattwthompson:
I got too fancy with this, I'm going to split it into something more incremental since this is merging ~4 tools into one

Issue #311, opened 2025-01-07
Implement geomeTRIC option `--converge maxiter`


amcisaac:
We'd like to use the GeomeTRIC option `--converge maxiter` (as documented [here](https://geometric.readthedocs.io/en/latest/options.html#optimization-parameters)), which allows the program to exit gracefully after hitting the maximum number of iterations, but it's not currently supported in QCSubmit. We will open a PR to implement it as one of the options for `convergence_set` in `GeometricProcedure`.

amcisaac:
This is currently blocked by the fact that `--converge maxiter` is (as far as I can tell) not implemented in a release version of GeomeTRIC, documented in [this issue](https://github.com/leeping/geomeTRIC/issues/213)

j-wags:
Copying an email I sent earlier to complete the records here:

> We'd be happy to have a PR for this and run it on our stack/compute, but I wonder if we can avoid basically testing in prod when we submit to QCArchive (this has led to problems in the past... like I wouldn't put it past QCF to silently fail to pass the keyword to geometric or something weird like that). So could you 
> 
> 1. add a test to the PR with a super short opt of a tiny mol, or that exports the task to the MolSSI QC ecosystem and ensures that the correct keyword is set in the real dataset
> 2. Manually run 5 or 10 of these step-limited opts on a QCFractal server locally to ensure that they have the desired behavior
> 
> I just checked and @amcisaac's still in the developers group on GitHub, so feel free to work directly in a branch of QCSubmit and use the PR to run CI. Once this is merged we can make a stable release for the provenance and any qca-dataset-submission stuff that needs the new version.

amcisaac:
Thanks for copying that, I should have included it! We will absolutely include tests both locally and in the CI.

PR #314, opened 2025-02-01
Add custom convergence options and `--converge maxiter` to `GeometricProcedure`

amcisaac:
## Description
Implementing custom convergence criteria in `GeometricProcedure`, to allow users to either specify their own convergence criteria, or use the `--converge maxiter` option to allow the program to exit gracefully on the maximum number of iterations. Implements #311 

I implemented this by adding a new keyword for `GeometricProcedure`, `converge`, which accepts a list of strings corresponding to the convergence flags accepted by GeomeTRIC. I tested this in QCEngine to ensure that it should propagate correctly. I also added new validators to ensure that the `converge` options are valid and will run in GeomeTRIC, and ensure no clashes between `convergence_set` and `converge`. 

Unfortunately the keyword names are a bit confusing in GeomeTRIC, with `--converge maxiter` allowing it to exit gracefully upon hitting the max number of iterations, and `--maxiter X` specifying the actual max number of iterations. I tried to make the documentation around this as clear as possible, but please let me know if there are improvements that can be made.

## Todos
  - [x] Implementing custom convergence criteria via the `converge` keyword
  - [x] Implementing support for the `--converge maxiter` flag, which can be used either with a regular `convergence_set` or with custom convergence criteria
  - [x] Locally test functions to ensure correct behavior 
  - [x] Locally test resulting keyword block in QCEngine GeomeTRIC calculations
  - [x] Write unit tests to ensure correct behavior in QCArchive

## Questions
- [ ] I wasn't able to figure out a good way to validate both the two `converge` and `convergence_set` fields in `GeometricProcedure`. The best option I could find was to use a `root_validator` to do it, but I wasn't able to figure out how to import that from the `validators.py` script, so now it's a bit messy. If anyone has suggestions for how to clean this up, please let me know.

## Status
- [x] Ready to go


amcisaac:
I _think_ this PR should be ready to go, but currently the CI tests are failing due to #315 . I also had some tests fail a few days ago due to plotly requiring the `anywidget` module to be available (per this [error in my CI report](https://github.com/openforcefield/openff-qcsubmit/actions/runs/13082891619/job/36509787871)). I can make these changes (remove `memory_cache_key` and require `anywidget` in the conda envs), but they didn't necessarily seem like small changes and I thought folks may want to discuss them.

Additionally, we need to add a new config file for Read The Docs (per[ this error message](https://readthedocs.org/projects/openff-qcsubmit/builds/27065569/)), but I'm not sure what the details are of our read the docs build. I'm happy to work on this, but need some guidance on how to proceed.

amcisaac:
Also very happy for any feedback on the code style or implementation, I tried to follow the structure already present but I'm not sure if I succeeded!

openff/qcsubmit/validators.py
```diff
@@ -39,6 +39,55 @@ def literal_upper(literal: str) -> str:
...
+    for i, keyword in enumerate(convergence_keyword_list):
+
+        # Check if the entry is in the allowed keys
+        if keyword.lower() in allowed_keys:
+
+            # maxiter must not be followed by a number
+            if keyword.lower() == "maxiter":
+                if i != len(convergence_keyword_list) - 1:
+                    if convergence_keyword_list[i + 1].lower() not in allowed_keys:
+                        raise AssertionError(
```
jthorton, line 59: I think all of the `AssertionError`s need a test against them, it might also be better to make some new exceptions for these and add them [here](https://github.com/openforcefield/openff-qcsubmit/blob/main/openff/qcsubmit/exceptions.py) I think being specific with errors is better in this case.

openff/qcsubmit/_tests/test_submissions.py
```diff
@@ -835,6 +838,139 @@ def test_optimization_submissions(fulltest_client, specification):
...
+        assert record.error is None
+
+        assert [
+            key.lower() for key in record.specification.keywords["converge"]
+        ] == dataset.optimization_procedure.converge
+
+        # Length of trajectory is the number of steps. Should be equal to maxiter + 1
+        # if --converge maxiter was requested
+        if "maxiter" in converge:
+            assert len(record.trajectory) == dataset.optimization_procedure.maxiter + 1
```
jthorton, line 966: Nice this is great!

openff/qcsubmit/_tests/test_submissions.py
```diff
@@ -835,6 +838,139 @@ def test_optimization_submissions(fulltest_client, specification):
...
+    index = ethane.to_smiles()
+    dataset.add_molecule(index=index, molecule=ethane)
+
+    # force a validation error with the GeometricProcedure
+    with pytest.raises(ValidationError):
+        dataset.optimization_procedure = GeometricProcedure(
+            convergence_set="GAU maxiter"
+        )
+    with pytest.raises(ValidationError):
+        dataset.optimization_procedure = GeometricProcedure(converge=["GAU", "maxiter"])
```
jthorton, line 925: It might be best to move these to their own tests, that should make it easier to work out what has happened if this test failed as currently it could be caused by three different cases.

openff/qcsubmit/_tests/test_submissions.py
```diff
@@ -521,6 +523,7 @@ def test_adding_compute(fulltest_client, dataset_data):
     await_services(fulltest_client, max_iter=30)
 
     # make sure of the results are complete
+    print(dataset.type)
```
jthorton, line 526: test print

openff/qcsubmit/_tests/test_submissions.py
```diff
@@ -835,6 +838,139 @@ def test_optimization_submissions(fulltest_client, specification):
...
+        legacy_qcsubmit_ds_type_to_next_qcf_ds_type[dataset.type], dataset.dataset_name
+    )
+
+    query = ds.iterate_records(specification_names="test_spec")
+
+    for name, spec, record in query:
+        assert record.status == RecordStatusEnum.complete
+        assert record.error is None
+
+        assert [
```
jthorton, line 959: Nice, so this checks that the custom convergence settings are passed to the record. Could we also check they are picked up by geometric in the stdout of the optimisation?

jthorton (commented):
Thanks @amcisaac this looks great and will be very useful! I have few comments my main ask is to test all of the new validation introduced.

openff/qcsubmit/validators.py
amcisaac: Pydantic's validators only let you raise `ValueError`, `AssertionError`, or a sub-class of those. So I created two new, more descriptive errors, but I wasn't able to make them `QCSubmitException`s. I can also revert it back if we want all our exceptions to be `QCSubmitException`s

amcisaac:
Thanks so much for your suggestions! I think I've addressed everything, let me know what you think.

mattwthompson:
I took the liberty of adding a few minor formatting fixes and patching over minimal dependency issues to probably get CI passing

openff/qcsubmit/exceptions.py
```diff
@@ -217,3 +217,21 @@ class RecordTypeError(QCSubmitException):
 
     error_type = "record_type_error"
     header = "Record Type Error"
+
+
+class InvalidConvergeSettingsError(AssertionError):
```
mattwthompson, line 222:
```suggestion
class InvalidConvergeSettingsError(QCSubmitException):
```

openff/qcsubmit/exceptions.py
```diff
@@ -217,3 +217,21 @@ class RecordTypeError(QCSubmitException):
...
+    """
+    Invalid converge options were passed to GeometricProcedure()
+    """
+
+    error_type = "invalid_converge_settings_error"
+    header = "Invalid Converge Settings Error"
+
+
+class ConflictingConvergeSettingsError(ValueError):
```
mattwthompson, line 231:
```suggestion
class ConflictingConvergeSettingsError(QCSubmitException):
```

Just going by what else in in this class, I think you meant to inherit this way

mattwthompson (approved):
So `--maxiter` must include a number (int-like?) but `--converge maxiter` must _not_ 🫠

LGTM % good test coverage. I only understand this functionality as described here and the linked issues but I think it succeeds in doing what was scoped out, again assuming that the changed lines are hit by tests

Process of merge/release is @j-wags's responsibility as project owner so I will defer there

openff/qcsubmit/exceptions.py
amcisaac: According to the docs, a Pydantic validator can only raise a `ValueError` or `AssertionError`, not any other type of error. @JoshHorton suggested that I should make custom exceptions for these cases, rather than just use the basic `ValueError` or `AssertionError`, but I _think_ that means that they can't inherit from the `QCSubmitException`. If we prefer all custom exceptions to inherit from `QCSubmitException`, I can just revert these to generic `ValueError` or `AssertionError`.

mattwthompson: Ah yeah I see that now. I'd suggest either
* inheriting from both (not sure the proper order) if it works
* simply dropping the `error_type` and `header` lines, since they probably won't get used if not inheriting from the custom type and I think they're only used for pretty-ifying the error message

amcisaac:
Okay great, thanks so much, and thank you for fixing the tests! Yes, unfortunatley `--maxiter` must take an int but `--converge maxiter` takes nothing 😅 

Looks like one line isn't being tested (and I thought it was...). I can fix that

mattwthompson:
@amcisaac let us know if you wanted to tinker with anything more, otherwise I think this is good to go and I'll hit the button

jthorton (approved):
Thanks @amcisaac this is great!

amcisaac:
I'm happy with it, if you don't mind the way the exceptions are structured! Sorry, haven't had a time to mess around with inheriting two types

PR #332, opened 2025-03-05
Use `openmmforcefields` 0.14.2

### LINKED ISSUE #331: Remove Interchange constraint in test environments
This is blocked by `openmmforcefields` 0.14.2

mattwthompson:
## Description
Resolves #331 - should bring down 0.4.2

## Todos
  - [x] Nudge the solver to use `openmmforcefields` 0.14.2
  - [x] Remove Interchange constraint
 
## Status
- [x] Ready to go

### DIFF
--- a/devtools/conda-envs/basic.yaml
+++ b/devtools/conda-envs/basic.yaml
@@ -39,8 +39,8 @@ dependencies:
   # this should be changed to openff-toolkit-base
 
   - openff-toolkit
-  - openff-interchange-base <0.4.2  # 0.4.2 conflicts with openmmforcefields
-  - openff-units ~=0.2.1
+  - openff-interchange-base
+  - openff-units
   - pydantic
   - pyyaml
   - torsiondrive
@@ -49,5 +49,5 @@ dependencies:
   - h5py>=3.6.0
 
   # Optional
-  - openmmforcefields
+  - openmmforcefields ~=0.14.2
   - openff-fragmenter-base >=0.2.0
--- a/devtools/conda-envs/psi4.yaml
+++ b/devtools/conda-envs/psi4.yaml
@@ -36,8 +36,8 @@ dependencies:
   ### Core dependencies.
 
   - openff-toolkit-base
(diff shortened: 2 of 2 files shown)


devtools/conda-envs/basic.yaml
```diff
@@ -49,5 +49,5 @@ dependencies:
   - h5py>=3.6.0
 
   # Optional
-  - openmmforcefields
+  - openmmforcefields ~=0.14.2
```
j-wags, line 52: (not blocking) slight preference, otherwise this effectively becomes a downpin against 0.14.3. The patch I put into OMMFFs should make it compatible with both the old and new interchange behavior. 

```suggestion
  - openmmforcefields >=0.14.2
```

devtools/conda-envs/psi4.yaml
```diff
@@ -47,7 +47,7 @@ dependencies:
   - h5py>=3.6.0
 
   # Optional
-  - openmmforcefields >=0.9.0
+  - openmmforcefields ~=0.14.2
```
j-wags, line 50: (not blocking) ditto
```suggestion
  - openmmforcefields >=0.14.2
```

j-wags (approved):

devtools/conda-envs/basic.yaml
mattwthompson: If I understand the standard[^1] correctly - and if conda follows `pip`'s behavior - this would allow 0.14.3 but _not_ 0.15.0. I read this as `>=0.14.2,<0.15.0`. My estimate is that 0.14.3 (through 0.14.9999999999....) would be less likely to include a behavior change than 0.15.0, same as a 1.0. I think this is what you're trying to get as well

[^1]: https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release

j-wags: Ah, you're right. I mistakenly thought the `~=` operator was dependent on the precision of the version specified (so `~=0.14` would allow up to `0.14.999`, but `~=0.14.2` would only allow up to `0.14.2.999`). I now see that's not the case, so you can disregard this comment :-)

PR #337, opened 2025-04-17
Error when torsion drive is probably still running

mattwthompson:
## Description
Fixes https://github.com/openforcefield/openff-bespokefit/issues/393 by raising a more useful error message.

Arguably this could be a warning instead of an error, but there is potential for a significant behavior change if i.e. a fit is run on a partially-complete torsion drive dataset and re-run later when the compute is complete. This is not necessarily wholly bad, but an opinionated change we did not want to make here.

(Some extra comments in the code, and much more context in the linked issue.)

## Status
- [x] Ready to go


j-wags (approved):
Awesome, thanks @mattwthompson!

PR #382, opened 2026-06-16
Update for QCF 0.65 and add mdtraj as testing dep

### LINKED ISSUE #380: CI failing
A few are from MDTraj no longer being packaged by the toolkit by default, easy fix

Most frequent comes from setup

```
../../../micromamba/envs/qcsubmit-test-psi4/lib/python3.12/site-packages/_pytest/config/__init__.py:1892: AttributeError

The above exception was the direct cause of the following exception:

pytestconfig = <_pytest.config.Config object at 0x7f188b797d70>

    @pytest.fixture(scope="function")
    def fulltest_client(pytestconfig):
        """
        A portal client used for full end-to-end tests
        """
    
>       uri = pytestconfig.getoption("--fractal-uri")
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       ValueError: no option named '--fractal-uri'
```

https://github.com/openforcefield/openff-qcsubmit/actions/runs/27245917977

### LINKED ISSUE #383: Docs builds are failing
We need to update the os version used 

Example: https://app.readthedocs.org/projects/openff-qcsubmit/builds/33170279/

```
 Config file validation error

Config validation error in build.os. Expected one of (ubuntu-22.04, ubuntu-24.04, ubuntu-26.04, ubuntu-lts-latest), got type str (ubuntu-20.04). Double check the type of the value. A string may be required (e.g. "3.10" instead of 3.10) 
```

j-wags:
- [x] Fix #380 
- [x] Fix #383

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,7 +1,7 @@
 version: 2
 
 build:
-  os: ubuntu-26.04
+  os: ubuntu-24.04
   tools:
     python: "mambaforge-latest"
   jobs:
@@ -20,5 +20,3 @@ sphinx:
   configuration: docs/conf.py
   fail_on_warning: false
 
-formats:
-- pdf
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -3,7 +3,10 @@ version: 2
 build:
   os: ubuntu-26.04
   tools:
-    python: "mambaforge-4.10"
+    python: "mambaforge-latest"
+  jobs:
+    pre_build:
+      - mamba list
 
 python:
   install:
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,7 +1,7 @@
 version: 2
 
 build:
-  os: ubuntu-20.04
+  os: ubuntu-26.04
   tools:
     python: "mambaforge-4.10"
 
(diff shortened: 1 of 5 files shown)


devtools/conda-envs/basic.yaml
```diff
@@ -25,7 +25,7 @@ dependencies:
 
   - qcengine >=0.25
   - qcelemental >=0.25.1
-  - qcfractal >=0.56
+  - qcfractal >=0.65
```
mattwthompson, line 28: (non-blocking) will we need to update this in packaging as well?

.readthedocs.yaml
```diff
@@ -1,9 +1,12 @@
...
 
 build:
-  os: ubuntu-20.04
+  os: ubuntu-24.04
   tools:
-    python: "mambaforge-4.10"
+    python: "mambaforge-latest"
+  jobs:
+    pre_build:
+      - mamba list
```
mattwthompson: (non-blocking) please remove if this was intended to be only for debugging. I can see it being generally useful to have, though, so not a problem to keep it

mattwthompson (approved):
LGTM

devtools/conda-envs/basic.yaml
j-wags: Maybe. I actually think the source code patch may be reverse compatible. I'll check this and will revert this change if so.

j-wags: Yes, I tried going down to the old version and tests failed, so I'll need to propagate this to packaging.

Issue #395, opened 2026-08-19
Update for QCFractal 0.70 


j-wags:
CI is [failing](https://github.com/openforcefield/openff-qcsubmit/actions/runs/32087458339/job/95562835557) now that it's pulling in QCFractal/QCPortal 0.70. ~It's possible that this is just due to QCF/QCP updating to pydantic v2 models and releasing before QCElemental and co get their [v2 updates](https://github.com/MolSSI/QCElemental/issues/323#issuecomment-2297528477) out as main-label packages.~ QCElemental has had pydantic v2 packages on their main label for a while.

bennybp:
Looks like a pydantic v2 change. I had fields marked as something like:

    class BaseRecord(BaseModel):
        ...
        properties: Optional[Dict[str, Any]]

Before, if properties wasn't given, it would default automatically to None. But that behavior changed:

https://pydantic.dev/docs/validation/latest/concepts/fields/#:~:text=In%20Pydantic%20V1%2C,in%20Pydantic%20V2.&text=Optional%20would%20be,in%20Pydantic%20V2.

So we could either fix the calls in qcsubmit, or I could provide the defaults in qcportal.

j-wags:
I'm not so familiar with pydantic and haven't looked at the business logic of QCSubmit for a while, so I'm going to try and wrap my head around this by seeing which changes in QCSubmit would be needed to get this working. Then I'll report back with an opinion on that choice :-)

bennybp:
I think I see a few places to get started if you want me to draft a PR

j-wags:
Go for it! I'll work on my own branch in parallel but will take from your branch in the (likely) case of my failure 😆

j-wags:
Making a new build of production package downpinning qcportal<0.70 [here](https://github.com/conda-forge/openff-qcsubmit-feedstock/pull/36)

j-wags:
Scoping this to just updating for new QCF. Migrating to pydantic v2 models can be done separately.

PR #396, opened 2026-08-19
Fix pydantic v2/qcportal v0.70 errors

bennybp:
## Description
Fixes errors related to qcportal v0.70 and newer versions of QCElemental, related to Pydantic v2 changes.

Not all are fixed yet, but working through them.

See Issue #395 

## Questions
- [ ] Is qcsubmit moving to pydantic v2?

## Status
- [ ] Ready to go


j-wags:
Sorry, this repo doesn't have an easy path for development without openeye. I've disabled the openeye setup, and while this will cause some tests to fail, the pydantic changes should be largely orthogonal (ie if the test suite passes except the OE dependent tests, it very likely passes)

